PR #2, opened 2024-04-23
Support setting up Openvswitch (Ovs) Bridge network

coiby:
Resolves: https://issues.redhat.com/browse/RHEL-33465

This patch supports setting up an Ovs bridge in kdump initrd. An Ovs
bridge is similar to a classic Linux bridge but we use ovs-vsctl to find
out the Ethernet device (having the MAC address as the bridge) added to
an Ovs bridge. Once we copy all the needed NetworkManager (NM) connection
profiles to kdump initrd and all the necessary files, NM will create an Ovs bridge
automatically in kdump initrd.

In the case of OpenShift Container Platform (OCP),
ovs-configuration.service [1] is responsible for setting up an Ovs bridge.
In theory, we can also try to bring up the original physical network
interface before ovs-configuration.service. But this approach is
cumbersome because it breaks our assumption that we should bring up the
same network in kdump intrd as in 1st kernel (establishing the same network
in kdump initrd only needs to copy the needed NM connection profiles
thus we don't need to learn how different network setup work under the
hood).

```
How to test this patch with the help of configure-ovs.sh?
=========================================================

1. Extract configure-ovs.sh from [2]

2. Install necessary packages for configure-ovs.sh
    dnf install openvswitch -yq
    dnf install NetworkManager-ovs nmap-ncat -yq

    systemctl enable --now openvswitch

    # restart NM so the ovs plugin can be activated
    systemctl restart NetworkManager

3. Assume the network interface used for creating an Ovs bridge is
   "ens2", use configure-ovs.sh to create an Ovs bridge,

    interface=ens2
    mkdir -p /etc/ovnk
    echo $interface > /etc/ovnk/iface_default_hint
    bash configure-ovs.sh OVNKubernetes

4. (Optional) If you want to make the created Ovs bridge survive a
   reboot, simply make the created NM connections created by
   configure-ovs.sh persist,

    cp /run/NetworkManager/system-connections/ovs-* /etc/NetworkManager/system-connections/

If you need to create an Ovs bridge on top of a bonding network, use the
following commands for step 3,

    nmcli con add type bond ifname bond0
    nmcli con add type ethernet ifname eth0 master bond0
    nmcli con add type ethernet ifname eth1 master bond0

    echo bond0 > /etc/ovnk/iface_default_hint
    bash configure-ovs.sh OVNKubernetes
```

Note
1. For RHEL, openvswitch3.3 may be installed so we need to get the
   package name by "rpm -qf /usr/lib/systemd/system/openvswitch.service"
2. For RHEL9, openvswitch package needs to installed from another repo,
    cat << 'EOF' > /etc/yum.repos.d/ovs.repo
    [rhosp-rhel-9-fdp-cdn]
    name=Red Hat Enterprise Linux Fast Datapath $releasever - $basearch cdn
    baseurl=http://rhsm-pulp.corp.redhat.com/content/dist/layered/rhel9/$basearch/fast-datapath/os/
    enabled=1
    gpgcheck=0
    EOF

    dnf install openvswitch3.3 -yq

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/units/ovs-configuration.service.yaml
[2] https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml

### DIFF
--- a/dracut-module-setup.sh
+++ b/dracut-module-setup.sh
@@ -478,6 +478,24 @@ kdump_setup_vlan() {
     _save_kdump_netifs "$_parent_netif"
 }
 
+kdump_setup_ovs() {
+    local _netdev="$1"
+    local _dev _phy_if
+
+    _phy_if=$(ovs_find_phy_if "$_netdev")
+
+    if kdump_is_bridge "$_phy_if"; then
+        kdump_setup_vlan "$_phy_if"
+    elif kdump_is_bond "$_phy_if"; then
+        kdump_setup_bond "$_phy_if" || return 1
+    elif kdump_is_team "$_phy_if"; then
+        derror "Ovs bridge over team is not supported!"
+        exit 1
+    fi
+
+    _save_kdump_netifs "$_phy_if"
+}
+
 # setup s390 znet
 kdump_setup_znet() {
     local _netif
@@ -513,6 +531,28 @@ kdump_get_remote_ip() {
     echo "$_remote"
 }
 
+# Find the physical interface of Open vSwitch (Ovs) bridge
+#
+# The physical network interface has the same MAC address as the Ovs bridge
+ovs_find_phy_if() {
+    local _mac _dev
+    _mac=$(kdump_get_mac_addr $1)
+
+    for _dev in $(ovs-vsctl list-ifaces $1); do
+        if [[ $_mac == $(</sys/class/net/$_dev/address) ]]; then
+            echo -n "$_dev"
+            return
+        fi
+    done
+
+    return 1
+}
+
+# Tell if a network interface is an Open vSwitch (Ovs) bridge
+kdump_is_ovs_bridge() {
+    [[ $(_get_nic_driver $1) == openvswitch ]]
+}
+
 # Collect netifs needed by kdump
 # $1: destination host
 kdump_collect_netif_usage() {
@@ -536,6 +576,9 @@ kdump_collect_netif_usage() {
         kdump_setup_team "$_netdev"
     elif kdump_is_vlan "$_netdev"; then
         kdump_setup_vlan "$_netdev"
+    elif kdump_is_ovs_bridge "$_netdev"; then
+        has_ovs_bridge=yes
+        kdump_setup_ovs "$_netdev"
     fi
     _save_kdump_netifs "$_netdev"
 
@@ -582,6 +625,29 @@ kdump_install_resolv_conf() {
     fi
 }
 
+kdump_install_ovs_deps() {
+    [[ $has_ovs_bridge == yes ]] || return 0
+    inst_multiple -o $(rpm -ql NetworkManager-ovs) $(rpm -ql $(rpm -qf /usr/lib/systemd/system/openvswitch.service)) /sbin/sysctl /usr/bin/uuidgen /usr/bin/hostname /usr/bin/touch /usr/bin/expr /usr/bin/id /usr/bin/install /usr/bin/setpriv /usr/bin/nice /usr/bin/df
+    # 1. Overwrite the copied /etc/sysconfig/openvswitch so
+    # ovsdb-server.service can run as the default user root.
+    # /etc/sysconfig/openvswitch by default intructs ovsdb-server.service to
+    # run as USER=openvswitch, However openvswitch doesn't have the permission
+    # to write to /tmp in kdump initrd and ovsdb-server.servie will fail
+    # with the error "ovs-ctl[1190]: ovsdb-server: failed to create temporary
+    # file (Permission denied)". So run ovsdb-server.service as root instead
+    #
+    # 2. Bypass the error "referential integrity violation: Table Port column
+    # interfaces row" caused by we changing the connection profiles
+    echo "OPTIONS=\"--ovsdb-server-options='--disable-file-column-diff'\"" >"${initdir}/etc/sysconfig/openvswitch"
+
+    KDUMP_DROP_IN_DIR="${initdir}/etc/systemd/system/nm-initrd.service.d"
+    mkdir -p "$KDUMP_DROP_IN_DIR"
+    printf "[Unit]\nAfter=openvswitch.service\n" >$KDUMP_DROP_IN_DIR/01-after-ovs.conf
+
+    $SYSTEMCTL -q --root "$initdir" enable openvswitch.service
+    $SYSTEMCTL -q --root "$initdir" add-wants basic.target openvswitch.service
+}
+
 # Setup dracut to bring up network interface that enable
 # initramfs accessing giving destination
 kdump_install_net() {
@@ -595,6 +661,7 @@ kdump_install_net() {
         kdump_install_nm_netif_allowlist "$_netifs"
         kdump_install_nic_driver "$_netifs"
         kdump_install_resolv_conf
+        kdump_install_ovs_deps
     fi
 }
 
@@ -1006,6 +1073,7 @@ EOF
 
 install() {
     declare -A unique_netifs ipv4_usage ipv6_usage
+    local has_ovs_bridge
 
     kdump_module_init
     kdump_install_conf


coiby:
Hi @jbtrystram, can you take a look at this patch set? Thank you! Btw, if this PR gets merged, https://github.com/openshift/machine-config-operator/pull/4213 will no longer be needed.

dracut-module-setup.sh
```diff
@@ -502,6 +502,29 @@ kdump_get_remote_ip() {
     echo "$_remote"
 }
 
+# Find the physical interface of Openvswitch (Ovs) bridge
+#
+# The physical network interface has the same MAC address as the Ovs bridge
+ovs_find_phy_if() {
+    local _mac _dev
+    _mac=$(</sys/class/net/$1/address)
```
prudo1, line 529: There's a kdump_get_mac_addr function doing the same thing. Although it's currently unused...

dracut-module-setup.sh
```diff
@@ -327,11 +327,30 @@ _install_nmconnection() {
     inst "$_src" "$_dst"
 }
 
+kdump_install_ovs_nmconnection() {
+    local _uuid _uuid _cloned_nm_path _netif
```
prudo1, line 331: _uuid is unused and defined twice

dracut-module-setup.sh
```diff
@@ -327,11 +327,30 @@ _install_nmconnection() {
...
+        return 0
+    done <<<"$(nmcli --get-values filename,ACTIVE connection show | grep "no$")"
+}
+
 kdump_install_nmconnections() {
     local _netif _nm_conn_path _cloned_nm_path
 
     while IFS=: read -r _netif _nm_conn_path; do
         [[ -v "unique_netifs[$_netif]" ]] || continue
+        [[ $ovs_phy_if == $_netif ]] && continue
```
prudo1, line 353: Do you really need a separate kdump_install_ovs_nmconnection? The way I see it kdump_install_nmconnection and kdump_install_ovs_nmconnection are very similar. Wouldn't it be enough  to have

`is_ovs_bridge "$_netif" && _netif=$(ovs_find_phy_if "$_netif"`

here? That would require a few more calls to is_ovs_bridge but would remove the need to have the extra function and the global variable ovs_phy_if.

prudo1 (commented):
Hi Coiby,

personally I would merge the two commits as I don't see much value in having the first one. But I don't have a strong opinion on that.

If possible I would like to simplify the second commit. Having ovs_phy_if as global variable and the extra kdump_install_ovs_nmconnection function seems over complicated to me on first view.

Thanks
Philipp

dracut-module-setup.sh
coiby: Good catch!

dracut-module-setup.sh
coiby: The connection profile we try to bring up in kdump initrd is actually inactive in 1st kernel, so I still think we need an extra function.

Let's me give an concrete example. For the following case, currently the kdump dracut module finds br-ex is used for network communication with remote fs,
```
sh-5.1# nmcli c show         
NAME                UUID                                  TYPE           DEVICE 
ovs-if-br-ex        e9b9247b-f3e3-4f8b-9890-969aff471eff  ovs-interface  br-ex  
br-ex               118c4996-d3b3-4350-b4b2-b16fd0e5a6eb  ovs-bridge     br-ex  
ovs-if-phys0        569596f1-69f7-4e6c-8a88-21614291ab38  ethernet       ens5   
ovs-port-br-ex      a3290363-e7f8-4d4b-ab13-c5cdae8dc121  ovs-port       br-ex  
ovs-port-phys0      141f408a-27fc-4638-a188-276ad9d0f043  ovs-port       ens5   
Wired connection 1  778a339a-4809-32c3-9abc-cc7c82ebe094  ethernet       --     
``` 
But we actually want to bring up the last inactive profile "Wired connection 1" instead.

jbtrystram:
> Hi @jbtrystram, can you take a look at this patch set? Thank you! 
TBH i don't know this code at all so a review would be unhelpful.. 

> Btw, if this PR gets merged, [openshift/machine-config-operator#4213](https://github.com/openshift/machine-config-operator/pull/4213) will no longer be needed.
Thanks for letting me know ! I'll update the MCO afterwards then

dracut-module-setup.sh
prudo1: Hi Coiby,
you are right. My suggestion didn't take into account that we also need to get the .nmconnection for the inactive connection, which is exactly what kdump_install_ovs_nmconnection does. So we need to keep it...

Nevertheless could you move the call of kdump_install_ovs_nmconnection into kdump_install_nmconnection rather than calling it in kdump_install_net? In my opinion that makes it clearer that this is a special case for this function. Plus it should allow to get rid off the ovs_phy_if global variable.

In particular I mean this change (haven't tested it)

```
diff --git a/dracut-module-setup.sh b/dracut-module-setup.sh
index 9374548..a568489 100755
--- a/dracut-module-setup.sh
+++ b/dracut-module-setup.sh
@@ -328,10 +328,9 @@ _install_nmconnection() {
 }
 
 kdump_install_ovs_nmconnection() {
-    local _uuid _uuid _cloned_nm_path _netif
+    local _cloned_nm_path _netif
 
-    [[ -n $ovs_phy_if ]] || return
-    _netif=$ovs_phy_if
+    _netif=$(ovs_find_phy_if "$1")
 
     while IFS=: read -r _nm_conn_path _; do
         [[ $(nmcli --get-values connection.interface-name connection show "$_nm_conn_path") == $_netif ]] || continue
@@ -350,7 +349,12 @@ kdump_install_nmconnections() {
 
     while IFS=: read -r _netif _nm_conn_path; do
         [[ -v "unique_netifs[$_netif]" ]] || continue
-        [[ $ovs_phy_if == $_netif ]] && continue
+
+       if is_ovs_bridge "$_netif"; then
+               kdump_install_ovs_nmconnection "$_netif"
+               continue
+       fi
+
         if _cloned_nm_path=$(clone_and_modify_nmconnection "$_netif" "$_nm_conn_path"); then
             _install_nmconnection "$_cloned_nm_path"
         else
@@ -569,11 +573,6 @@ kdump_collect_netif_usage() {
         kdump_setup_vlan "$_netdev"
     fi
 
-    if is_ovs_bridge "$_netdev"; then
-        ovs_phy_if=$(ovs_find_phy_if "$_netdev")
-        _netdev=$ovs_phy_if
-    fi
-
     _save_kdump_netifs "$_netdev"
 
     if [[ ! -f ${initdir}/etc/cmdline.d/50neednet.conf ]]; then
@@ -627,7 +626,6 @@ kdump_install_net() {
     _netifs=$(_get_kdump_netifs)
     if [[ -n "$_netifs" ]]; then
         kdump_install_nmconnections
-        kdump_install_ovs_nmconnection
         apply_nm_initrd_generator_timeouts
         kdump_setup_znet "$_netifs"
         kdump_install_nm_netif_allowlist "$_netifs"
@@ -1049,7 +1047,7 @@ remove_cpu_online_rule() {
 
 install() {
     declare -A unique_netifs ipv4_usage ipv6_usage
-    local arch ovs_phy_if
+    local arch
 
     kdump_module_init
     kdump_install_conf
```

coiby:
> Hi Coiby,

Hi Phillipp,

> 
> personally I would merge the two commits as I don't see much value in having the first one. But I don't have a strong opinion on that.

Previously, I used two commits because it may be easier for the OpenShift team to review. But since [configure-ovs.sh](https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml) contains the info I have, now I squash two commits together in the new version.

This new version switches to a new approach i.e. we now try to bring up the same network in kdump initrd as in the 1st kernel. The reason is I find a Ovs bridge can be created on top of another network like bonding network. And it's cumbersome to support this case using the previous approach.

> 
> If possible I would like to simplify the second commit. Having ovs_phy_if as global variable and the extra kdump_install_ovs_nmconnection function seems over complicated to me on first view.
> 
> Thanks Philipp

dracut-module-setup.sh
coiby: Thanks for the suggestion! As I switch to another approach (as explained to [the second to last comment](https://github.com/rhkdump/kdump-utils/pull/2#issuecomment-2090628732)), the suggestion no longer applies. The new approach still needs a global variable but the code is more modular and easier to review.

jbtrystram:
Anything blocking that PR ?

kenneth-dsouza:
Hello team, any update here?

daveyoung (commented):
Hi Coiby, can you explain more about why this is needed? Have you encounter a bug? what is the steps and symptom? 

Also rethink about the original issue,  dracut should be able to handle those dns issues automatically, if no it sounds better to fix in dracut. 

Can you split this one to another pr if this is not ovs specific so that the review can be separately.

coiby:
Hi @jbtrystram and @kenneth-dsouza, 

Currently, the progress is blocked by code review. It seems you Cores team asks someone else to review patches about Openvswitch network. Can I ask this person to help this PR as well?

prudo1 (approved):

jbtrystram:
> Currently, the progress is blocked by code review. It seems you Cores team asks someone else to review patches about Openvswitch network. Can I ask this person to help this PR as well?

I am not sure who you are referring to ? There is no harm in asking for a review !

dracut-module-setup.sh
```diff
@@ -564,13 +608,34 @@ kdump_install_resolv_conf() {
...
     inst "$_resolv_conf"
     if NetworkManager --print-config | grep -qs "^dns=none"; then
         printf "[main]\ndns=none\n" > "${initdir}/${_nm_conf_dir}"/90-dns-none.conf
     fi
 }
 
+kdump_install_ovs_deps() {
+    [[ $has_ovs_bridge == yes ]] || return 0
+    inst_multiple -o $(rpm -ql NetworkManager-ovs) $(rpm -ql openvswitch) /usr/bin/uuidgen /usr/bin/hostname /usr/bin/touch /usr/bin/expr /usr/bin/id /usr/bin/install /usr/bin/setpriv /usr/bin/nice /usr/bin/df
+    # By default, ovsdb-server.service runs as USRE=openvswitch, However
```
mkp-rh, line 620: Nit: USER

dracut-module-setup.sh
```diff
@@ -564,13 +608,34 @@ kdump_install_resolv_conf() {
...
+    # openvswitch doesn't have the permission to write to /tmp in kdump initrd
+    # and ovsdb-service.servie will failure with the error
+    # "ovs-ctl[1190]: ovsdb-server: failed to create temporary file (Permission denied)"
+    # So run ovsdb-server.service as root instead
+    mkdir -p "${initdir}/etc/sysconfig/system.conf.d"
+    echo "root:root" >"${initdir}/etc/sysconfig/openvswitch"
```
mkp-rh, line 626: Shouldn't this be `OVS_USER_ID="root:root"`?

One other option to running OVS as root could be to set TMPDIR to a writable directory, if one exists.

JM1:
@coiby Could you please explain, how your code sets up the Open vSwitch bridges in the initrd? What bridges get created? What OCP release did you test your changes with?

JM1:
Where is `ovs-system` (in `kdump_collect_netif_usage()`) coming from?

How does your code restore OCP's `br-ex` bridge?

dracut-module-setup.sh
coiby: Thanks for catching this typo!

dracut-module-setup.sh
coiby: Yes, `OVS_USER_ID="root:root"` is what I meant. But your comment made me realize my understanding is incorrect because `echo "root:root" >"${initdir}/etc/sysconfig/openvswitch"` also works. So actually ovsdb-server.service by default run as root unless /etc/sysconfig/openvswitc intructs ovsdb-server.service to run as openvswitch.

dracut-module-setup.sh
```diff
@@ -564,13 +608,34 @@ kdump_install_resolv_conf() {
...
+    echo "OPTIONS=\"--ovsdb-server-options='--disable-file-column-diff'\"" >>"${initdir}/etc/sysconfig/openvswitch"
+
+    KDUMP_DROP_IN_DIR="${initdir}/etc/systemd/system/nm-initrd.service.d"
+    mkdir -p $KDUMP_DROP_IN_DIR
+    printf "[Unit]\nAfter=openvswitch.service\n" >$KDUMP_DROP_IN_DIR/01-after-ovs.conf
+
+    $SYSTEMCTL -q --root "$initdir" enable openvswitch.service
+    $SYSTEMCTL -q --root "$initdir" add-wants basic.target openvswitch.service
+}
+
```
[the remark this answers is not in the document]
coiby: Hi @prudo1! Yes, it's better to move it to dracut or even this kdump dracut module in the long term. Unfortunately, this may need more code-refactoring than  we have expected. In order for our kdump dracut module tell dracut to include the future ovs dracut module, we need to move the logic of detecting if it's an OVS bdrige to either the check function or mkdumprd.

coiby:
> @coiby Could you please explain, how your code sets up the Open vSwitch bridges in the initrd? What bridges get created? What OCP release did you test your changes with?

Hi @JM1! Thanks for raising this question. The code simply copy the NetworkManager connection files and the supporting files to kdump initrd and then NM will set up the same Open vSwitch bdrige to the 1st kernel automatically. I forgot which OCP version I tested early. But later I realize this not limited to OCP so I test it in RHEL9 and Fedora40. If you are interested in more details, I've included the info about how I set up a testing environment in the commit message.

> 
> What did `nmcli -t -f device,filename connection show --active` (in `kdump_install_nmconnections()`) return in your tests?

```
ovs-if-br-ex        de17dd7e-dd41-4865-a99e-60077f0837ae  ovs-interface  br-ex  
br-ex               c5662992-05c0-4b05-9302-a554c705ef6a  ovs-bridge     br-ex  
ovs-if-phys0        ddcae60d-7d24-46b2-86bc-c326b6688cc9  ethernet       eno1   
ovs-port-br-ex      3ffa5d8b-5ae6-45c6-9fd6-04e5f4b2eeaa  ovs-port       br-ex  
ovs-port-phys0      aadfdac7-fb7f-4aa0-9180-0eb64263f722  ovs-port       eno1   
```
Btw, I've also tested the case of  Open vSwitch bridge over a bonding network.

coiby:
> Where is `ovs-system` (in `kdump_collect_netif_usage()`) coming from?

I can't find `ovs-system`. Or do you mean the `kdump_is_ovs_bridge` function?

> How does your code restore OCP's `br-ex` bridge?

This is done by NM. I simply copy all the needed NetworkManager connection files and the supporting files to kdump initrd.

dracut-module-setup.sh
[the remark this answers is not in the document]
coiby: Hi @jbtrystram! How soon do you expect [RHEL-33465](https://issues.redhat.com/browse/RHEL-33465) to be resolved? It's ideal to have an upstream ovs dracut module but it will takes much more time. So we have a short term fix firstly if that's what you prefer.

JM1:
> > Where is `ovs-system` (in `kdump_collect_netif_usage()`) coming from?
> 
> I can't find `ovs-system`. Or do you mean the `kdump_is_ovs_bridge` function?
> 

Your previous [commit](https://github.com/rhkdump/kdump-utils/compare/471c0d942d535a656871aeb62b7694977814779c..eb6c659fac86a77414b22b098df4690119186315) had this line:
```
_save_kdump_netifs "ovs-system"
```
But it is gone in later commits. Hence, solved by removal 😄

> > How does your code restore OCP's `br-ex` bridge?
> 
> This is done by NM. I simply copy all the needed NetworkManager connection files and the supporting files to kdump initrd.

Ok. If I understand your code correctly, in `kdump_collect_netif_usage` it will collect the netdev (`br-ex`) that would be used when accessing the remote fs (e.g. via ssh). It would also identify the corresponding physical NIC such as `enp1s0` with the help of `kdump_setup_ovs` and `ovs_find_phy_if` and add both (`br-ex` and `enp1s0`) to `unique_netifs`. Later, in `kdump_install_nmconnections` it would retrieve all NM connections associated with those two devices. I did not test it but the approach seems reasonable. 

However, I am wondering why you do not simply try to copy and restore all NetworkManager connections? Wouldn't that reduce the amount of code here dramatically?

coiby:
> > > Where is `ovs-system` (in `kdump_collect_netif_usage()`) coming from?
> > 
> > 
> > I can't find `ovs-system`. Or do you mean the `kdump_is_ovs_bridge` function?
> 
> Your previous [commit](https://github.com/rhkdump/kdump-utils/compare/471c0d942d535a656871aeb62b7694977814779c..eb6c659fac86a77414b22b098df4690119186315) had this line:
> 
> ```
> _save_kdump_netifs "ovs-system"
> ```
> 
> But it is gone in later commits. Hence, solved by removal 😄

Oh my poor memory! This reminds me to do another test on OCP. Or if you think there is no need to do it, please let me know.

> 
> > > How does your code restore OCP's `br-ex` bridge?
> > 
> > 
> > This is done by NM. I simply copy all the needed NetworkManager connection files and the supporting files to kdump initrd.
> 
> Ok. If I understand your code correctly, in `kdump_collect_netif_usage` it will collect the netdev (`br-ex`) that would be used when accessing the remote fs (e.g. via ssh). It would also identify the corresponding physical NIC such as `enp1s0` with the help of `kdump_setup_ovs` and `ovs_find_phy_if` and add both (`br-ex` and `enp1s0`) to `unique_netifs`. Later, in `kdump_install_nmconnections` it would retrieve all NM connections associated with those two devices. I did not test it but the approach seems reasonable.
> 
> However, I am wondering why you do not simply try to copy and restore all NetworkManager connections? Wouldn't that reduce the amount of code here dramatically?

Good suggestion! `--ovsdb-server-options='--disable-file-column-diff'` also prompt me to think if I can simply copy all active NetworkManager connections to initrd at once. But there are three cases (currently solutions need changing the NM connection profiles) needing further investigation and different solutions,
- Some environments like some Azure machine doesn't use persistent NIC name (the current solution is to modify a NM connection profile to match a device by MAC address, for details check commit 568623e69a32266a3b00225b9de6a93435c44474) 
- If a NIC has an IPv4 or IPv6 address, set the corresponding may-fail
    property to no. Otherwise, dumping vmcore over IPv6 could fail because
    only IPv4 network is ready or vice versa (currently solution is to disable IPv6 if only IPv4 is used and vice versa)
- Some NICs need longer connection.wait-device-timeout otherwise the connection will fail to be established (commit 6b586a903632831b05f591048fdb7de4af273088).

JM1:
>... if I can simply copy all active NetworkManager connections to initrd at once. But there are three cases (currently solutions need changing the NM connection profiles) needing further investigation and different solutions,
> 
> * Some environments like some Azure machine doesn't use persistent NIC name (the current solution is to modify a NM connection profile to match a device by MAC address, for details check commit [568623e](https://github.com/rhkdump/kdump-utils/commit/568623e69a32266a3b00225b9de6a93435c44474))
> 
> * If a NIC has an IPv4 or IPv6 address, set the corresponding may-fail  property to no. Otherwise, dumping vmcore over IPv6 could fail because only IPv4 network is ready or vice versa (current solution is to disable IPv6 if only IPv4 is used and vice versa, for details check commit [9dfcacf](https://github.com/rhkdump/kdump-utils/commit/9dfcacf72d2cd135bf8b72c4bbc97f15efa21250))
> 
> * Some NICs need longer connection.wait-device-timeout otherwise the connection will fail to be established (commit [6b586a9](https://github.com/rhkdump/kdump-utils/commit/6b586a903632831b05f591048fdb7de4af273088)).

Please, document this ^ in the commit message (or your code). You put a lot of thoughts into the why's ( :+1: :+1: :+1: ), so help us or a future engineer with understanding your reasoning.

I am not sure if copying [configure-ovs.sh](https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml) to initrd and executing it there (maybe wrapped somehow), might not still be less hassle in the long run. But your approach also works (i assume, did not test it), so please go ahead and merge ☺️

coiby:
> > ... if I can simply copy all active NetworkManager connections to initrd at once. But there are three cases (currently solutions need changing the NM connection profiles) needing further investigation and different solutions,
> > 
> > * Some environments like some Azure machine doesn't use persistent NIC name (the current solution is to modify a NM connection profile to match a device by MAC address, for details check commit [568623e](https://github.com/rhkdump/kdump-utils/commit/568623e69a32266a3b00225b9de6a93435c44474))
> > * If a NIC has an IPv4 or IPv6 address, set the corresponding may-fail  property to no. Otherwise, dumping vmcore over IPv6 could fail because only IPv4 network is ready or vice versa (current solution is to disable IPv6 if only IPv4 is used and vice versa, for details check commit [9dfcacf](https://github.com/rhkdump/kdump-utils/commit/9dfcacf72d2cd135bf8b72c4bbc97f15efa21250))
> > * Some NICs need longer connection.wait-device-timeout otherwise the connection will fail to be established (commit [6b586a9](https://github.com/rhkdump/kdump-utils/commit/6b586a903632831b05f591048fdb7de4af273088)).
> 
> Please, document this ^ in the commit message (or your code). You put a lot of thoughts into the why's ( 👍 👍 👍 ), so help us or a future engineer with understanding your reasoning.
> 

I've documented it the commit message! Thanks for the excellent suggestion!

> I am not sure if copying [configure-ovs.sh](https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml) to initrd and executing it there (maybe wrapped somehow), might not still be less hassle in the long run. But your approach also works (i assume, did not test it), so please go ahead and merge ☺️

Thanks for reviewing the patches and endorsing this approach! As for running configure-ovs.sh in kdump initrd, configure-ovs.sh assumes there is an active NM connection first. For kdump initrd, this active NM doesn't exist (on the other hand if it exists, there is no need to use configure-ovs.sh).

dracut-module-setup.sh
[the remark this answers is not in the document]
jbtrystram: @coiby thanks for the ping !  Having a short term fix and a tracking issue to make sure we adjust when the dracut module lands is fine with me ! 
Thanks for the work

daveyoung:
@JM1  hi, can you have a look at the  comment  https://github.com/rhkdump/kdump-utils/issues/30#issuecomment-2419339354
Or if you know who is the right person to review the ovs stuff please let us know, thanks!  I tried to tag you in the issue, but failed.

JM1:
@drizzt is our OVS packaging wizard :mage: He knows much better! :rocket: 

Could you please have a look at @daveyoung's comment above, @drizzt? ☺️

PR #10, opened 2024-06-08
kdump LUKS support

coiby:
LUKS is the standard for Linux disk encryption. Many users choose LUKS
and in some use cases like Confidential VM it's mandated. With kdump
enabled, when the 1st kernel crashes, the system could boot into the
kdump/crash kernel and dump the memory image i.e. /proc/vmcore to a
specified target. Currently, when dumping vmcore to a LUKS
encrypted device, there are two problems,

 - Kdump kernel may not be able to decrypt the LUKS partition. For some
   machines, a system administrator may not have a chance to enter the
   password to decrypt the device in kdump initramfs after the 1st kernel
   crashes; For cloud confidential VMs, depending on the policy the
   kdump kernel may not be able to unseal the keys with TPM and the
   console virtual keyboard is untrusted.

 - LUKS2 by default use the memory-hard Argon2 key derivation function
   which is quite memory-consuming compared to the limited memory reserved
   for kdump. Take Fedora example, by default, only 256M is reserved for
   systems having memory between 4G-64G. With LUKS enabled, ~1300M needs
   to be reserved for kdump. Note if the memory reserved for kdump can't
   be used by 1st kernel i.e. an user sees ~1300M memory missing in the
   1st kernel.

Based on the new kernel feature that dm-crypt keys can persist for the
kdump kernel [1][2], this patch which is adapted from [3]
1) ask the 1st kernel to save a copy of the LUKS volume keys
2) ask the kdump kernel to add the copy of the LUKS volume keys to
   specified keyring and then use --volume-key-keyring the unlock the
   LUKS device.

Users need to set up the link-volume-key option in /etc/crypttab (man
crypttab) so kdumpctl can read the key from @u::%logon:cryptsetup:UUID
and instruct the kernel to save it to kdump-reserved memory.

[1] https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/kdump/kdump.rst#write-the-dump-file-to-encrypted-disk-volume
[2] https://github.com/torvalds/linux/commit/180cf31af7c313790f1e0fba1c7aa42512144dd5
[3] https://lists.fedorahosted.org/archives/list/kexec@lists.fedoraproject.org/message/Y3KUSJQPN3JHUUC2FPIK7H4HTSX2TUCX/

### DIFF
--- a/kdumpctl
+++ b/kdumpctl
@@ -1105,6 +1105,9 @@ prepare_luks()
 		return 1
 	fi
 
+	# For the case of CPU/memory hotplugging, we can reuse loaded keys
+	[[ $(cat $LUKS_CONFIGFS/reuse) == 1 ]] && return 0
+
 	for _luks_dev in "${_luks_devs[@]}"; do
 		_devuuid=$(maj_min_to_uuid "$_luks_dev")
 		_key_dir=$LUKS_CONFIGFS/$_devuuid
--- a/kdumpctl
+++ b/kdumpctl
@@ -1056,9 +1056,42 @@ check_final_action_config()
 	esac
 }
 
+_get_luks_key_by_unlock()
+{
+	local _devuuid=$1 _key_des=$2
+	local _max_retries
+	local _attempt=1
+
+	local _luks_unlock_cmd=""
+
+	# Check if stdin is a terminal.
+	if [ -t 0 ]; then
+		_max_retries=5
+		ddebug "Attempting to unlock LUKS device. You have $_max_retries attempts."
+	else
+		# Not a terminal (e.g., running as system service), so only try once
+		# for cases where volume key is sealed to TPM which doesn't need user
+		# interaction.
+		_max_retries=1
+		ddebug "Attempting to unlock LUKS device (non-interactive mode)..."
+	fi
+
+	while [ "$_attempt" -le "$_max_retries" ]; do
+		if cryptsetup open "UUID=$_devuuid" DUMMY "--link-vk-to-keyring=@u::%logon:$_key_des" --test-passphrase; then
+			ddebug "Success: LUKS device unlocked."
+			dwarn "To avoid manually running kdumpctl, ensure the link-volume-key=@u::%logon:$_key_des option is correctly set up in /etc/crypttab (see man crypttab)."
+			return 0
+		fi
+		_attempt=$((_attempt + 1))
+	done
+
+	derror "Error: Could not unlock the LUKS device."
+	return 1
+}
+
 prepare_luks()
 {
-	local _luks_dev _key_id _key_des
+	local _luks_dev _key_id _key_des _luks_unlock_cmd
 	declare -a _luks_devs
 
 	mapfile -t _luks_devs < <(get_all_kdump_crypt_dev)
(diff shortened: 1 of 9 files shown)


kdumpctl
```diff
@@ -1030,6 +1032,53 @@ check_final_action_config()
...
+	if [[ $_state == fresh ]]; then
+		printf "init %s" "$_count" > $LUKS_SYSFS
+	fi
+
+	for _luks_dev in "${_luks_devs[@]}"; do
+		_devuuid=$(maj_min_to_uuid "$_luks_dev")
+		printf "record cryptsetup:%s" "$_devuuid" > $LUKS_SYSFS
+	done
+}
+
```
prudo1, line 1081: There's no need to use `printf` here. A simple `echo "init $_count"` and `echo "record cryptsetup:$(maj_min_to_uuid "$_luks_dev")"` should work just fine.
Plus `_devuuid` isn't defined local.

kdumpctl
```diff
@@ -1030,6 +1032,53 @@ check_final_action_config()
...
+		return 0
+	fi
+}
+
+prepare_luks()
+{
+	local _luks_dev _count _cmds _cmd _state
+	declare -a _luks_devs
+
+	_state=$(< $LUKS_SYSFS)
```
prudo1, line 1054: What happens when `$LUKS_SYSFS` doesn't exist? E.g. when you are running on an old kernel or one where the required configs aren't set?

dracut-module-setup.sh
```diff
@@ -1004,6 +1004,55 @@ ForwardToConsole=yes
...
+            printf -- 'RUN+="/sbin/initqueue  --settled --unique --onetime '
+            printf -- '--name kdump-crypt-target-%%k %s ' "$(command -v cryptsetup)"
+            printf -- 'luksOpen --volume-key-keyring %s $env{DEVNAME} %s"\n' "%%user:$_key_desc" "luks-$_devuuid"
+        } >>"${initdir}/etc/udev/rules.d/70-luks-kdump.rules"
+
+
+    done
+
+    echo >"$initdir/etc/cmdline.d/90crypt.conf"
+    echo >"$initdir/etc/crypttab"
```
prudo1, line 1044: Do I understand it correctly, that you are undoing work other dracut modules have done? If so, could you please add a short comment so we still remember it in a couple of years.

prudo1 (commented):
Hi @coiby,
As far as I can tell the overall code looks fine. I've got a few questions and nits to improve though.
Thanks
Philipp

kdumpctl
```diff
@@ -1030,6 +1032,53 @@ check_final_action_config()
 	esac
 }
 
+LUKS_SYSFS=/sys/kernel/crash_dm_crypt_keys
+reuse_loaded_luks_keys()
+{
+	local _luks_dev _count _cmds _cmd _state
+	declare -a _luks_devs
+
```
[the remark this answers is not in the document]
prudo1: On second thought, wouldn't it be safer to simply drop this function and always setup the a new key ring when loading the kdump kernel. It's more expensive but otherwise you could end up in a situation like
```
# dump to luks dev1
$ kdumpctl start
# edit kdump.conf to dump to luks dev2
$ kdumpctl rebuild
$ kdumpctl reload
```
which, IIUC, should fail as the key of dev1 is now used to decrypt dev2, which  most likely should fail.

kdumpctl
```diff
@@ -1052,9 +1052,38 @@ check_final_action_config()
...
 
+_get_luks_key_by_unlock() {
+    local _unlock_cmd="$1"
+    local _max_retries
+    local _attempt=1
+
+    # Check if stdin is a terminal.
+    if [ -t 0 ]; then
+        _max_retries=5
+        ddebug "Attempting to unlock LUKS device. You have $max_retries attempts."
```
licliu, line 1063: max_retries should be _max_retries

licliu (commented):
Hi @coiby, this patch set looks good to me, just find 2 typos.

kdumpctl
```diff
@@ -1052,9 +1052,38 @@ check_final_action_config()
...
+    else
+	# Not a terminal (e.g., running as system service), so only try once
+	# for cases where volume key is sealed to TPM which doesn't need user
+	# interaction.
+	_max_retries=1
+        ddebug "Attempting to unlock LUKS device (non-interactive mode)..."
+    fi
+
+    while [ "$_attempt" -le "$_max_retries" ]; do
+        if eval "$_unlock_cmd"; then
```
licliu, line 1073: I don't like eval, it is better if we can avoid it.

licliu:
A small nitpick about the naming of $LUKS_CONFIGFS/reuse and $LUKS_CONFIGFS_REUSE, but otherwise it looks good. However, due to the merge of another PR, some conflicts have appeared and need to be resolved.

licliu (approved):
Hi @coiby, this version looks good to me.I give my approval to this PR, but if you'd like to make changes based on the comments in the open gemini thread, I'd be happy to take another look.

coiby:
Hi @licliu,

Thanks for reviewing and approving the PR! I'll merge PR with small improvement suggested by Gemini i.e. "the $(command -v cryptsetup) part can be precomputed and stored in a variable to avoid re-evaluation in each loop iteration".

For other Gemini comments, I believe I've gone through them. If there is one I need to pay particular attention, please let me know! I'll address them in later PRs. Thanks!

PR #13, opened 2024-06-18
[PATCH v2]Introduce vmcore creation notification to kdump

liutgnu:
Motivation
    ==========
    
    unexpected for kdump.
    
    It is highly recommended people to recheck kdump after any system
    modification, such as:
    
       firmware upgrading etc.
       etc.
    
    Design
    ======
    
    indicator of such modification, and kdump need to be rechecked. This will
    clear the vmcore creation status specified in $VMCORE_CREATION_STATUS.
    
    Vmcore creation check will happen at "kdumpctl (re)start/status", and will
    report the creation success/fail status to users. A "success" status indicates
    previously there has been a vmcore successfully generated based on the current
    env, so it is more likely a vmcore will be generated later when real crash
    happens; A "fail" status indicates previously there was no vmcore
    generated, or has been a vmcore creation failed based on current env. User
    should check the 2nd kernel log or the kexec-dmesg.log for the failing reason.
    
    the current env. The format will be like:
    
       success 1718682002
    
    Which means, there has been a vmcore generated successfully at this
    timestamp for the current env.
    
    Usage
    =====
    
    kdump: Stopping kdump: [OK]
    kdump: Starting kdump: [OK]
    kdump: Fail: No vmcore creation test performed!
    
    [root@localhost ~]# echo c > /proc/sysrq-trigger
    
    kdump: Kdump is operational
    kdump: Success: Last successful vmcore creation on Tue Jun 18 16:39:10 CST 2024
    
    kdump: Stopping kdump: [OK]
    kdump: Starting kdump: [OK]
    kdump: Success: Last successful vmcore creation on Tue Jun 18 16:39:10 CST 2024
    
    The notification for kdumpctl (re)start/status can be disabled by
    setting VMCORE_CREATION_NOTIFICATION in /etc/sysconfig/kdump
    
### DIFF
--- a/dracut-kdump.sh
+++ b/dracut-kdump.sh
@@ -237,16 +237,8 @@ save_opalcore_fs()
 	return 0
 }
 
-dump_to_rootfs()
+try_mount_sysroot()
 {
-
-	if [ "$(systemctl status dracut-initqueue | sed -n "s/^\s*Active: \(\S*\)\s.*$/\1/p")" = "inactive" ]; then
-		dinfo "Trying to bring up initqueue for rootfs mount"
-		systemctl start dracut-initqueue
-	fi
-
-	dinfo "Clean up dead systemd services"
-	systemctl cancel
 	dinfo "Waiting for rootfs mount, will timeout after 90 seconds"
 	systemctl start --no-block sysroot.mount
 
@@ -258,6 +250,23 @@ dump_to_rootfs()
 
 	if ! is_mounted /sysroot; then
 		derror "Failed to mount rootfs"
+		return 1
+	fi
+	return 0
+}
+
+dump_to_rootfs()
+{
+
+	if [ "$(systemctl status dracut-initqueue | sed -n "s/^\s*Active: \(\S*\)\s.*$/\1/p")" = "inactive" ]; then
+		dinfo "Trying to bring up initqueue for rootfs mount"
+		systemctl start dracut-initqueue
+	fi
+
+	dinfo "Clean up dead systemd services"
+	systemctl cancel
+
+	if ! try_mount_sysroot; then
 		return
 	fi
 
@@ -312,11 +321,19 @@ do_final_action()
 
 do_dump()
 {
+	if ! try_mount_sysroot; then
+		return 1
+	fi
+	set_vmcore_creation_status 'clear' "/sysroot/$VMCORE_CREATION_STATUS"
+
 	eval $DUMP_INSTRUCTION
 	_ret=$?
 
 	if [ $_ret -ne 0 ]; then
+		set_vmcore_creation_status 'fail' "/sysroot/$VMCORE_CREATION_STATUS"
 		derror "saving vmcore failed"
+	else
+		set_vmcore_creation_status 'success' "/sysroot/$VMCORE_CREATION_STATUS"
 	fi
 
 	return $_ret
--- a/gen-kdump-sysconfig.sh
+++ b/gen-kdump-sysconfig.sh
@@ -47,6 +47,10 @@ KDUMP_IMG="vmlinuz"
 #What is the images extension.  Relocatable kernels don't have one
 KDUMP_IMG_EXT=""
 
+# Enable vmcore creation notification by default, disable by setting
+# VMCORE_CREATION_NOTIFICATION=""
+VMCORE_CREATION_NOTIFICATION="yes"
+
 # Logging is controlled by following variables in the first kernel:
 #   - @var KDUMP_STDLOGLVL - logging level to standard error (console output)
 #   - @var KDUMP_SYSLOGLVL - logging level to syslog (by logger command)
--- a/kdump-lib-initramfs.sh
+++ b/kdump-lib-initramfs.sh
@@ -9,6 +9,7 @@ KDUMP_CONFIG_FILE="/etc/kdump.conf"
 FENCE_KDUMP_CONFIG_FILE="/etc/sysconfig/fence_kdump"
 FENCE_KDUMP_SEND="/usr/libexec/fence_kdump_send"
 LVM_CONF="/etc/lvm/lvm.conf"
+VMCORE_CREATION_STATUS="/var/crash/vmcore-creation.status"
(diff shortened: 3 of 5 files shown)


baoquan-he (commented):
Hi Tao,

Thanks for the effort for making this achievement. I have one tiny concern about this PR.

If vmcore was created successfully and the status is recorded. Then later kdump kernel didn't finish booting to the init process or run the kdump scripts. Will we still get a 'success' status for kdump?

liutgnu:
> Hi Tao,
> 
> Thanks for the effort for making this achievement. I have one tiny concern about this PR.
> 
> If vmcore was created successfully and the status is recorded. Then later kdump kernel didn't finish booting to the init process or run the kdump scripts. Will we still get a 'success' status for kdump?

Hi Baoquan,

Yes, it will get a 'success' status. Because there was a vmcore generated before, so there is a success status recorded. Since we don't make any change to the system, the status file will not be deleted, and we are expecting the same env will lead to a same result, aka a vmcore generated for a later crash. However when a later crash happens, the init doesn't bootup. This breaks our premise of 
      "A 'success' status indicates previously there has been a vmcore successfully generated based on the current env, so 
     it is more likely a vmcore will be generated later when real crash happens". 

This is a problem and it cannot be solved by this patch, the patch is based on the premise, but the premise is failed.

Personally I suggest we give more time of testing the premise. If it cannot cover the majority proportion of our cases, there is no need to introduce this patch.

daveyoung (commented):
kdump: Stopping kdump: [OK]
kdump: Starting kdump: [OK]
kdump: Fail: No vmcore creation test performed!

Above "Fail" message causes misunderstandings, people may think that kdump service failed.  As it is just a notification, not mandatory remove the "Fail:" and "Success:"  looks reasonable.   And also extra brief words to let people know how to do the test will be better.  For example: kdump: it seems kdump not tested, you can trigger crash and test kdump via "echo c > /proc/sysrq-trigger" .    Maybe see if Philipp has better suggestion about the wording.

liutgnu:
Hi Dave,

Thanks for the review.

> [root@localhost ~]# kdumpctl restart kdump: kexec: unloaded kdump kernel kdump: Stopping kdump: [OK] kdump: kexec: loaded kdump kernel kdump: Starting kdump: [OK] kdump: Fail: No vmcore creation test performed!
> 
> Above "Fail" message causes misunderstandings, people may think that kdump service failed. As it is just a notification, not mandatory remove the "Fail:" and "Success:" looks reasonable. And also extra brief words to let people know how to do the test will be better. For example: kdump: it seems kdump not tested, you can trigger crash and test kdump via "echo c > /proc/sysrq-trigger" . Maybe see if Philipp has better suggestion about the wording.

I updated the notify message as:

    kdump: Stopping kdump: [OK]
    kdump: Starting kdump: [OK]
    kdump: Vmcore creation notice: No vmcore creation test performed!

The Fail/Success message been changed to "Vmcore creation notice". Personally I think the "kdump: it seems kdump not tested, you can trigger crash and test kdump via "echo c > /proc/sysrq-trigger"" message is redundant. Because the message is only meaningful to the CEE people, so CEE people know how to handle the test IMHO. And we don't want to instruct ordinary people to follow the message, because crash & reboot the system is a dangerous operation if people have no context about this feature. So I guess one "No vmcore creation test performed!" message is enough. What do you think?

Thanks,
Tao Liu

liutgnu:
Please checkout the new version of v3 in [1]. And I will close this one. Thanks!

[1]: https://github.com/rhkdump/kdump-utils/pull/20

baoquan-he:
Hi Tao,
>...
> Yes, it will get a 'success' status. Because there was a vmcore generated before, so there is a success status recorded. Since we don't make any change to the system, the status file will not be deleted, and we are expecting the same env will lead to a same result, aka a vmcore generated for a later crash. However when a later crash happens, the init doesn't bootup. This breaks our premise of "A 'success' status indicates previously there has been a vmcore successfully generated based on the current env, so it is more likely a vmcore will be generated later when real crash happens".
> 
When I talked to Dave about this, an idea comes up that we can use a specific command to trigger the test, e.g kdumpctl test, then remove the file $VMCORE_CREATION_STATUS before triggering crash. Then we will only have  'success' or 'unknown'/'untested' two status. 'success' means vmcore is saved successfully last time, 'untested/unknown' means it's not tested or vmcore is not saved successfully. Not sure if this is clearer regarding to the status.

> This is a problem and it cannot be solved by this patch, the patch is based on the premise, but the premise is failed.
> 
> Personally I suggest we give more time of testing the premise. If it cannot cover the majority proportion of our cases, there is no need to introduce this patch.

liutgnu:
Hi Baoquan,
> Hi Tao,
> 
> > ...
> > Yes, it will get a 'success' status. Because there was a vmcore generated before, so there is a success status recorded. Since we don't make any change to the system, the status file will not be deleted, and we are expecting the same env will lead to a same result, aka a vmcore generated for a later crash. However when a later crash happens, the init doesn't bootup. This breaks our premise of "A 'success' status indicates previously there has been a vmcore successfully generated based on the current env, so it is more likely a vmcore will be generated later when real crash happens".
> 
> When I talked to Dave about this, an idea comes up that we can use a specific command to trigger the test, e.g kdumpctl test, then remove the file $VMCORE_CREATION_STATUS before triggering crash. Then we will only have 'success' or 'unknown'/'untested' two status. 'success' means vmcore is saved successfully last time, 'untested/unknown' means it's not tested or vmcore is not saved successfully. Not sure if this is clearer regarding to the status.

Thanks for the comments, now I have worked out the v4 which implemented the testing entrance as "kdumpctl test", which will remove $VMCORE_CREATION_STATUS before triggering crash. Please checkout the final force-pushed commits in [1].

[1]: https://github.com/rhkdump/kdump-utils/pull/20
> 
> > This is a problem and it cannot be solved by this patch, the patch is based on the premise, but the premise is failed.
> > Personally I suggest we give more time of testing the premise. If it cannot cover the majority proportion of our cases, there is no need to introduce this patch.

PR #16, opened 2024-06-25
dracut-module-setup: Remove remove_cpu_online_rule() since PowerPC us…

pfliu:
dracut-module-setup: Remove remove_cpu_online_rule() since PowerPC uses nr_cpus
    
    Now, PowerPC uses nr_cpus cmdline option, there is no worry about
    bringing up all possible cpu since normally we finally only have 8 or 16
    cpus if nr_cpus=1 on PowerPC platform. So 40-redhat.rules will not raise
    trouble.
    
    Signed-off-by: Pingfan Liu <piliu@redhat.com>

### DIFF
--- a/dracut-module-setup.sh
+++ b/dracut-module-setup.sh
@@ -1004,31 +1004,12 @@ ForwardToConsole=yes
 EOF
 }
 
-remove_cpu_online_rule() {
-    local file=${initdir}/usr/lib/udev/rules.d/40-redhat.rules
-
-    if [[ -f $file ]]; then
-        sed -i '/SUBSYSTEM=="cpu"/d' "$file"
-    fi
-}
-
 install() {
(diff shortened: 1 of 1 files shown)


pfliu:
This is a clean up after using nr_cpus=1 on PowerPC platform
@sourabhjains, could you share your opinion

sourabhjains:
@pfliu Since `nr_cpus=1` is the default for PowerPC, I think we don't need to change `40-redhat.rules` for kdump initrd.

Thanks for the clean-up patch; the changes look good to me.

Note for users who manually fall back to `maxcpus=1`: make sure to adjust the crashkernel reservation to accommodate the additional memory required for more CPUs in the kdump environment.

coiby (approved):

PR #18, opened 2024-06-26
Various fixes

prudo1:
During the last test cycle and the following discussion several issues were found. This series fixes most of them.

Note: The last commit of the series depends on PR #12. Unfortunately I couldn't find a way to tell GitHub this connection. Thus this PR also contains the identical commit to the one from PR #12.

Depends on PR #12

PR #19, opened 2024-06-27
kdumpctl: Only update fadump for ppc64le 

licliu:
Fadump is only available for ppc64le, but reset_crashkernel() updates fadump
unconditionally, which will cause non-ppc64le architectures to always reboot
when run kdumpctl reset-crashkernel --reboot.

### DIFF
--- a/kdumpctl
+++ b/kdumpctl
@@ -1561,7 +1561,7 @@ reset_crashkernel()
 		_old_ck=$(get_grub_kernel_boot_parameter "$_kernel" crashkernel)
 		_new_ck=$(kdump_get_arch_recommend_crashkernel "$_new_dump_mode")
 		_old_fadump=$(get_grub_kernel_boot_parameter "$_kernel" fadump)
-		if _update_kernel_cmdline "$_kernel" fadump "$_old_fadump" "$_new_fadump"; then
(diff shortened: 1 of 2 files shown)


Barathrum-Liu:
I think this issue may be fixed by [kdumpctl: Simplify fadump handling in reset_crashkernel](https://github.com/rhkdump/kdump-utils/pull/18/commits/26fa48fa0a6517c2335faf685e7d4ab3fcfcd10b) in PR#18, if we decide to merge that PR, we can close this one.

prudo1:
> I think this issue may be fixed by [kdumpctl: Simplify fadump handling in reset_crashkernel](https://github.com/rhkdump/kdump-utils/pull/18/commits/26fa48fa0a6517c2335faf685e7d4ab3fcfcd10b) in [PR#18](https://github.com/rhkdump/kdump-utils/pull/18), if we decide to merge that PR, we can close this one.

@Barathrum-Liu: indeed that commit also fixes the problem this PR is supposed to solve. I've updated the commit message accordingly. In case there are no rejections I propose to use the commit from my PR as that also leads to cleaner code (at least in my opinion).

@licliu: I also added a commit to update the man page with slightly different wording compared to yours. I hope that is ok with you.

IMHO this PR can be closed.

PR #20, opened 2024-07-02
Introduce vmcore creation notification to kdump

liutgnu:
Introduce vmcore creation notification to kdump
    
    Motivation
    ==========
    
    unexpected for kdump.
    
    It is highly recommended people to recheck kdump after any system
    modification, such as:
    
       firmware upgrading etc.
       etc.
    
    Design
    ======
    
    indicator of such modification, and kdump need to be rechecked. This will
    clear the vmcore creation status specified in $VMCORE_CREATION_STATUS.
    
    Vmcore creation check will happen at "kdumpctl (re)start/status", and will
    report the creation success/fail status to users. A "success" status indicates
    previously there has been a vmcore successfully generated based on the current
    env, so it is more likely a vmcore will be generated later when real crash
    happens; A "fail" status indicates previously there was no vmcore
    generated, or has been a vmcore creation failed based on current env. User
    should check the 2nd kernel log or the kexec-dmesg.log for the failing reason.
    
    the current env. The format will be like:
    
       success 1718682002
    
    Which means, there has been a vmcore generated successfully at this
    timestamp for the current env.
    
    Usage
    =====
    
    kdump: Stopping kdump: [OK]
    kdump: Starting kdump: [OK]
    kdump: Notice: No vmcore creation test performed!
    
    [root@localhost ~]# kdumpctl test
    
    kdump: Kdump is operational
    kdump: Notice: Last successful vmcore creation on Tue Jun 18 16:39:10 CST 2024
    
    kdump: Stopping kdump: [OK]
    kdump: Starting kdump: [OK]
    kdump: Notice: Last successful vmcore creation on Tue Jun 18 16:39:10 CST 2024
    
    The notification for kdumpctl (re)start/status can be disabled by
    setting VMCORE_CREATION_NOTIFICATION in /etc/sysconfig/kdump
    
    ===
    
    v3 -> v2:
    Always mount
    $VMCORE_CREATION_STATUS(/var/crash/vmcore-creation.status)'s device for
    2nd kernel, in case /var is a seperate device than rootfs's device.
    
    v4 -> v3:
    Add "kdumpctl test" as the entrance for performing the kdump test.
    
    v5 -> v4:
    Fix the mounting failure issue in fadump.
    
    v6 -> v5:
    Add new argument as customized mount point for add_mount/to_mount.
    
    v7 -> v6:
    a. Code refactoring based on Philipp's suggestion.
    b. Only mount $VMCORE_CREATION_STATUS(/var/crash/vmcore-creation.status)'s
       device when needed.
    c. Add "--force" option for "kdumpctl test", to support the automation test
       script QE may perform.
    d. Add check in "kdumpctl test" that $VMCORE_CREATION_STATUS can only be on
       local drive.
    
    v8 -> v7:
    a. Rebased the patch on top of upstream commit e2b8463b5.
    b. Code refactoring based on Philipp's suggestion.
    c. Updated the "test" entry of kdumpctl.8.
    
### DIFF
--- a/kdump-lib-initramfs.sh
+++ b/kdump-lib-initramfs.sh
@@ -9,6 +9,7 @@ KDUMP_CONFIG_FILE="/etc/kdump.conf"
 FENCE_KDUMP_CONFIG_FILE="/etc/sysconfig/fence_kdump"
 FENCE_KDUMP_SEND="/usr/libexec/fence_kdump_send"
 LVM_CONF="/etc/lvm/lvm.conf"
+VMCORE_CREATION_STATUS="/var/crash/vmcore-creation.status"
 
 # Read kdump config in well formated style
 kdump_read_conf()
@@ -170,4 +171,36 @@ kdump_get_ip_route()
 kdump_get_ip_route_field()
 {
 	echo "$1" | sed -n -e "s/^.*\<$2\>\s\+\(\S\+\).*$/\1/p"
+}
+
+# $1: success/fail/clear
+# $2: status_file
+set_vmcore_creation_status()
+{
+	_status=$1
+	_status_file=$2
+	_dir=$(dirname "$_status_file")
+
+	[[ -d "$_dir" ]] || mkdir -p "$_dir"
+
+	_mnt_op=$(get_mount_info OPTIONS target "$_dir" -f)
+	case $_mnt_op in
+	ro*)
+		dinfo "remounting the vmcore status target in rw mode."
+		mount -o remount,rw "$(findmnt -n -o TARGET --target $_dir)"
+		;;
+	esac
+
+	case "$_status" in
+		success | fail)
+			dinfo "saving vmcore status file to $_status_file"
+			echo  "$_status $(date +%s)" > "$_status_file"
+			;;
+		clear)
+			rm -f "$_status_file"
+			;;
+		*)
+			return
+	esac
+	sync -f "$_dir"
 }
--- a/kdumpctl
+++ b/kdumpctl
@@ -185,6 +185,8 @@ rebuild_initrd()
 	else
 		rebuild_kdump_initrd
 	fi
+
+	set_vmcore_creation_status 'clear' "$VMCORE_CREATION_STATUS"
 }
 
 #$1: the files to be checked with IFS=' '
@@ -1043,6 +1045,7 @@ start()
 	start_dump || return
 
 	dinfo "Starting kdump: [OK]"
+	check_vmcore_creation_status
 	return 0
 }
 
@@ -1653,6 +1656,63 @@ _should_reset_crashkernel() {
 	[[ $(kdump_get_conf_val auto_reset_crashkernel) != no ]] && systemctl is-enabled kdump &> /dev/null
 }
 
+check_vmcore_creation_status()
+{
+	local _status _timestamp _status_date
+
+	[[ ${VMCORE_CREATION_NOTIFICATION,,} == "yes" ]] || return
+
+	if [[ ! -s $VMCORE_CREATION_STATUS ]]; then
+		dwarn "Notice: No vmcore creation test performed!"
+		return
+	fi
+
+	read -r _status _timestamp < "$VMCORE_CREATION_STATUS"
+	_status_date="$(date -d "@$_timestamp")"
+
+	if [[ "$_status" == "success" ]]; then
+		dinfo "Notice: Last successful vmcore creation on $_status_date"
+	else
+		dwarn "Notice: Last NOT successful vmcore creation on $_status_date"
+	fi
+}
+
+kdump_test()
+{
+	local _dir
+
+	if ! is_kernel_loaded "$DEFAULT_DUMP_MODE"; then
+		derror "Kdump needs be operational before test."
+		exit 1
+	fi
+
+	_dir=$(dirname "$VMCORE_CREATION_STATUS")
+	if ! [[ -d "$_dir" ]]; then
+		derror "Vmcore status dir $_dir not exist."
+		exit 1
+	fi
+
+	if ! lsblk $(get_mount_info SOURCE target "$_dir") > /dev/null; then
+		derror "$VMCORE_CREATION_STATUS must on local drive"
+		exit 1
+	fi
+
+	if [[ ! "$1" == "--force" ]]; then
+		read -p "DANGER!!! Will perform a kdump test by crashing the system, proceed? (y/N): " input
+		case $input in
+		[Yy] )
+			dinfo "Start kdump test..."
+			;;
+		* )
+			dinfo "Operation cancelled."
+			exit 0
+			;;
+		esac
+	fi
+	set_vmcore_creation_status 'clear' "$VMCORE_CREATION_STATUS"
+	echo c > /proc/sysrq-trigger
+}
+
 main()
 {
 	# Determine if the dump mode is kdump or fadump
@@ -1684,6 +1744,7 @@ main()
 			EXIT_CODE=3
 			;;
 		esac
+		check_vmcore_creation_status
 		exit $EXIT_CODE
 		;;
 	reload)
@@ -1728,8 +1789,12 @@ main()
 			reset_crashkernel_for_installed_kernel "$2"
 		fi
 		;;
+	test)
+		shift
+		kdump_test "$@"
+		;;
 	*)
-		dinfo $"Usage: $0 {estimate|start|stop|status|restart|reload|rebuild|reset-crashkernel|propagate|showmem}"
+		dinfo $"Usage: $0 {estimate|start|stop|status|restart|reload|rebuild|reset-crashkernel|propagate|showmem|test}"
 		exit 1
 		;;
 	esac
--- a/mkdumprd
+++ b/mkdumprd
@@ -73,9 +73,10 @@ has_dracut_module()
 # caller should ensure $1 is valid and mounted in 1st kernel
 to_mount()
 {
-	local _target=$1 _fstype=$2 _options=$3 _sed_cmd _new_mntpoint _pdev
+	local _target=$1 _fstype=$2 _options=$3 _new_mntpoint=$4
+	local _sed_cmd _pdev
 
-	_new_mntpoint=$(get_kdump_mntpoint_from_target "$_target")
+	_new_mntpoint="${_new_mntpoint:-$(get_kdump_mntpoint_from_target "$_target")}"
 	_fstype="${_fstype:-$(get_fs_type_from_target "$_target")}"
 	_options="${_options:-$(get_mntopt_from_target "$_target")}"
 	_options="${_options:-defaults}"
@@ -427,6 +428,17 @@ if ! is_fadump_capable; then
 	if fips-mode-setup --is-enabled 2> /dev/null; then
 		dracut_args+=(--add-device "$(findmnt -n -o SOURCE --target /boot)")
 	fi
+fi
+
+status_target=$(get_target_from_path $(dirname "$VMCORE_CREATION_STATUS"))
+
+if [[ $(get_root_fs_device) != "$status_target" ]]; then
+	new_mntpoint=$(echo /vmcorestatus/$(get_mntpoint_from_target "$status_target") \
+		| tr -s "/")
+	add_mount "$status_target" "" "" "$new_mntpoint"
+elif ! is_fadump_capable && \
+     ! [[ ${dracut_args[@]} == *"$(kdump_get_persistent_dev $status_target)"* ]]; then
+	add_mount "$status_target"
 fi
 
 dracut "${dracut_args[@]}" "$@"
--- a/dracut/99kdumpbase/kdump.sh
+++ b/dracut/99kdumpbase/kdump.sh
@@ -308,11 +308,22 @@ do_final_action() {
 }
 
 do_dump() {
+    if [ -d /vmcorestatus ]; then
+        _vmcore_creation_status="/vmcorestatus/$VMCORE_CREATION_STATUS"
+    else
+        _vmcore_creation_status="/sysroot/$VMCORE_CREATION_STATUS"
+    fi
+
+    set_vmcore_creation_status 'clear' "$_vmcore_creation_status"
+
     eval "$DUMP_INSTRUCTION"
     _ret=$?
 
     if [ $_ret -ne 0 ]; then
+        set_vmcore_creation_status 'fail' "$_vmcore_creation_status"
         derror "saving vmcore failed"
+    else
+        set_vmcore_creation_status 'success' "$_vmcore_creation_status"
     fi
 
     return $_ret
--- a/dracut/99kdumpbase/module-setup.sh
+++ b/dracut/99kdumpbase/module-setup.sh
@@ -1095,6 +1095,7 @@ install() {
     inst "/usr/bin/printf" "/sbin/printf"
     inst "/usr/bin/logger" "/sbin/logger"
     inst "/usr/bin/chmod" "/sbin/chmod"
+    inst "/usr/bin/dirname" "/sbin/dirname"
     inst "/lib/kdump/kdump-lib-initramfs.sh" "/lib/kdump-lib-initramfs.sh"
     inst "/lib/kdump/kdump-logger.sh" "/lib/kdump-logger.sh"
     inst "$moddir/kdump.sh" "/usr/bin/kdump.sh"
--- a/gen-kdump-sysconfig.sh
+++ b/gen-kdump-sysconfig.sh
@@ -47,6 +47,10 @@ KDUMP_IMG="vmlinuz"
 #What is the images extension.  Relocatable kernels don't have one
 KDUMP_IMG_EXT=""
 
+# Enable vmcore creation notification by default, disable by setting
+# VMCORE_CREATION_NOTIFICATION=""
+VMCORE_CREATION_NOTIFICATION="yes"
+
 # Logging is controlled by following variables in the first kernel:
 #   - @var KDUMP_STDLOGLVL - logging level to standard error (console output)
 #   - @var KDUMP_SYSLOGLVL - logging level to syslog (by logger command)
--- a/kdumpctl.8
+++ b/kdumpctl.8
@@ -70,6 +70,16 @@ Note: The memory requirements for kdump varies heavily depending on the
 used hardware and system configuration. Thus the recommended
 crashkernel might not work for your specific setup. Please test if
 kdump works after resetting the crashkernel value.
+.TP
+.I test [--force]
+Test the kdump by actually trigger the system crash & dump, and check if a
+vmcore can really be generated successfully based on current config and
+environment. After system reboot back to normal, check the test result
+by "kdumpctl status".
+
+If the optional parameter [--force] is provided, there will be no interact
+before triggering the system crash. Dangerous though, this option is meant
+for automation testing.
 
 .SH "SEE ALSO"
 .BR kdump.conf (5),


daveyoung:
Hi Tao,  otherwise I noticed that there is a special case for dump_to_rootfs in mkdumprd.  So doing some testing and verification with fadump should be also necessary.

liutgnu:
> Hi Tao, it is weird that the email notification still keeps using old description success/fail sentences. though it is V3 so no way to reply inline. Let me put a comment on web ui here: "kdump: Vmcore creation notice: No vmcore creation test performed!" For above notice, it looks redundant with two-time mentioning of "vmcore creation". I suggest just "notice: no vmcore creation test preformed!" Or "Warning: no vmcore creation test performed"
> 
> And "notice: Last successful vmcore creation on Tue Jun ..." or "Last successful vmcore creation on Tue Jun ..."
> 
> Also how about change "vmcore creation" to "kdump test", ditto about the status file name. It sounds more straightforward and easier to understand.

Hi Dave,

Agreed, just notice: Last successful vmcore .... is enough, will get it improved in v4.

liutgnu:
> Hi Tao, otherwise I noticed that there is a special case for dump_to_rootfs in mkdumprd. So doing some testing and verification with fadump should be also necessary.

Yes, will do it, currently there is a OOM issue in kexec-tools which blocked my fadump testing, see RHEL-35516.

coiby:
Hi @liutgnu,

I'm curious to ask why you prefer creating a new PR for a new version? For the perspective of maintainer, one benefit of pushing new version to the same PR is I can compare the difference between two versions with one-lick i.e. by clicking the "Compare" hyperlink, 
Screenshot_2024-07-02_15-54-56

liutgnu:
> Hi @liutgnu,
> 
> I'm curious to ask why you prefer creating a new PR for a new version? For the perspective of maintainer, one benefit of pushing new version to the same PR is I can compare the difference between two versions with one-lick i.e. by clicking the "Compare" hyperlink,

Please tell me which way you prefer. On one hand, the "force push" is dangerous and should be avoided, I guess a seperate MR of each patch version(pushing from different branch) is a safer one. On the other hand, "force push" against a single MR as you described, may have the following issue:

1: patch v1 (MR created)
 |
 V
2: force push1 (to improve v1)
 |
 V
3: force push2 (to improve v1 again, now the v1 is OK for review)
 |
 V
4: force push3 (Received comments, drafting v2)
 |
 V
5: force push4 (to improve v2, now v2 is ready for review)
 |
 V
....

You see, there might be plenty of "force pushes". I guess what you want to view is the diff of 3 & 5. I don't know if you can select the commit hash for diff comparison. What I know is it can only the diff of adjorning commits, please correct me if I'm wrong.

coiby:
Thanks for raising the concerns! If you are worried that a force-push may overwrite the history of this repo, then you needn't to worry about it because this repo doesn't allow force-push which is also the default setting.

As for the case of comparing versions of 3 & 5, you can simply send new version of patches in one single force-push to avoid this case. Btw, you can also compare different commits directly e.g. [compare your v3 with v2](https://github.com/rhkdump/kdump-utils/compare/7885a3e42fb32cf26f3a92c735adf8c91c0a2b6b..41454455f20cbe1e2498a898674c34295c9b74d0).

So I prefer sending new version of patches to the same PR. Another benefit is we can the same context in the same PR.

liutgnu:
The v4 has been force pushed to this PR, which mainly introduced the "kdumpctl test" as the entrance for performing kdump test.

In addition, in v3 Dave mentioned the combination of /kdumproot and /sysroot in kdump-lib.sh:get_kdump_mntpoint_from_target(), in order to simplify the dir mounting for saving $VMCORE_CREATION_STATUS. However I made an experiment showing the negative results, see the following:

If we unify the 2, and kdump through nfs and failure dump to rootfs, the mounting status will be like:

Arguments:  --add 'kdumpbase' --quiet --hostonly --hostonly-cmdline --hostonly-i18n --hostonly-mode 'strict' --hostonly-nics '' -o 'plymouth resume ifcfg earlykdump' --mount '192.168.1.65:/tmp/nfs /sysroot nfs defaults' --squash-compressor 'zstd' --mount '/dev/mapper/rhel_ibm--p8--kvm--03--guest--02-root /sysroot xfs rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota'

The original mounting status will be like:

Arguments:  --add 'kdumpbase' --quiet --hostonly --hostonly-cmdline --hostonly-i18n --hostonly-mode 'strict' --hostonly-nics '' -o 'plymouth resume ifcfg earlykdump' --mount '192.168.1.65:/tmp/nfs /kdumproot nfs defaults' --squash-compressor 'zstd' --mount '/dev/mapper/rhel_ibm--p8--kvm--03--guest--02-root /sysroot xfs rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota'

When nfs mount failed, it will try to mount sysroot as "systemctl start --no-block sysroot.mount", it will mount nfs again, instead of mounting the block device. So the mount will fail again, and dump_to_rootfs fails, which is unexpected.

see the 2nd kernel log:

[  130.266759] mount[563]: mount.nfs: Connection refused
[  130.268887] systemd[1]: sysroot.mount: Mount process exited, code=exited, status=32/n/a
[FAILED] Failed to mount /sysroot.
See 'systemctl status sysroot.mount' for details.
[  130.272324] systemd[1]: Failed to mount /sysroot.
...
[  256.480662] kdump[700]: failed to dump to '/sysroot', it's not a mount point!
[FAILED] Failed to start Kdump Vmcore Save Service.
See 'systemctl status kdump-capture.service' for details.
...
[  257.293836] systemd[1]: sysroot.mount: Directory /sysroot to mount over is not empty, mounting anyway.
[  257.296748] systemd[1]: Mounting /sysroot...
[  257.304670] mount[776]: mount.nfs: rpc.statd is not running but is required for remote locking.
...
[  347.314093] kdump[981]: Failed to mount rootfs
[  347.316864] kdump[983]: Executing final action systemctl reboot -f

daveyoung:
> The v4 has been force pushed to this PR, which mainly introduced the "kdumpctl test" as the entrance for performing kdump test. @daveyoung @baoquan-he
> 
> In addition, in v3 Dave mentioned the combination of /kdumproot and /sysroot in kdump-lib.sh:get_kdump_mntpoint_from_target(), in order to simplify the dir mounting for saving $VMCORE_CREATION_STATUS. However I made an experiment showing the negative results, see the following:
> 
> If we unify the 2, and kdump through nfs and failure dump to rootfs, the mounting status will be like:
> 
> Arguments: --add 'kdumpbase' --quiet --hostonly --hostonly-cmdline --hostonly-i18n --hostonly-mode 'strict' --hostonly-nics '' -o 'plymouth resume ifcfg earlykdump' --mount '192.168.1.65:/tmp/nfs /sysroot nfs defaults' --squash-compressor 'zstd' --mount '/dev/mapper/rhel_ibm--p8--kvm--03--guest--02-root /sysroot xfs rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota'
> 
> The original mounting status will be like:
> 
> Arguments: --add 'kdumpbase' --quiet --hostonly --hostonly-cmdline --hostonly-i18n --hostonly-mode 'strict' --hostonly-nics '' -o 'plymouth resume ifcfg earlykdump' --mount '192.168.1.65:/tmp/nfs /kdumproot nfs defaults' --squash-compressor 'zstd' --mount '/dev/mapper/rhel_ibm--p8--kvm--03--guest--02-root /sysroot xfs rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota'
> 
> When nfs mount failed, it will try to mount sysroot as "systemctl start --no-block sysroot.mount", it will mount nfs again, instead of mounting the block device. So the mount will fail again, and dump_to_rootfs fails, which is unexpected.
> 
> see the 2nd kernel log:
> 
> [ 130.266759] mount[563]: mount.nfs: Connection refused [ 130.268887] systemd[1]: sysroot.mount: Mount process exited, code=exited, status=32/n/a [FAILED] Failed to mount /sysroot. See 'systemctl status sysroot.mount' for details. [ 130.272324] systemd[1]: Failed to mount /sysroot. ... [ 256.480662] kdump[700]: failed to dump to '/sysroot', it's not a mount point! [FAILED] Failed to start Kdump Vmcore Save Service. See 'systemctl status kdump-capture.service' for details. ... [ 257.293836] systemd[1]: sysroot.mount: Directory /sysroot to mount over is not empty, mounting anyway. [ 257.296748] systemd[1]: Mounting /sysroot... [ 257.304670] mount[776]: mount.nfs: rpc.statd is not running but is required for remote locking. ... [ 347.314093] kdump[981]: Failed to mount rootfs [ 347.316864] kdump[983]: Executing final action systemctl reboot -f

Hi Tao, I finally found your reply here.  Thanks for trying the idea.  It is fair enough it you can not use one mount point.   Anyway I have still some questions about the mouting, added another comment in your new version

liutgnu:
Hi Dave, I just force pusted the v5 a few hours ago. I tested the fadump case, and mainly fixed its error mounting issues. Please checkout the added code comments in mkdumprd.

daveyoung (changes requested):
Hi, tao.   Let me put a new comment about the add_mount.

Can you do below?  

add a new argument $4 for mountpoint, eg. add_mount $1 $2 $3 $4
in add_mount() function you can choose the argument specified mount point, if $4 is empty then use the kdump mount point generated by another function.

In this way you do not need special handling about the notifier argument, you only specific the notifier mount point when you call add_mount

mkdumprd
```diff
@@ -407,6 +413,9 @@ if ! is_fadump_capable; then
 	# so it doesn't affect the logic of check_dump_fs_modified().
 	is_dump_to_rootfs && add_mount "$(to_dev_name "$(get_root_fs_device)")"
 
+	status_target=$(get_target_from_path $(dirname "$VMCORE_CREATION_STATUS"))
+	add_mount "$status_target" "" "" "vmcore_status_target"
+
```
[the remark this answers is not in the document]
liutgnu: Yes agreed, it would be better, I will get it improved in v6. Thanks!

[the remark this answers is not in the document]
liutgnu: Hi Dave, I force pusted v6 to this MR to fix the above issue, please check it out. Thanks!

kdumpctl
```diff
@@ -1729,8 +1758,27 @@ main()
...
+		[Yy] )
+			dinfo "Start kdump test..."
+			;;
+		* )
+			dinfo "Operation cancelled."
+			exit 0
+			;;
+		esac
+		set_vmcore_creation_status 'clear' "$VMCORE_CREATION_STATUS"
+		echo c > /proc/sysrq-trigger
```
prudo1, line 1778: I'd prefer if you moved this into a separate function rather than having it all in the switch case.

kdumpctl
```diff
@@ -1729,8 +1758,27 @@ main()
 			reset_crashkernel_for_installed_kernel "$2"
 		fi
 		;;
+	test)
+		is_kernel_loaded "$DEFAULT_DUMP_MODE"
+		if [[ "$?" -ne 0 ]]; then
+			derror "Kdump should be operational before test."
+			exit 1
+		fi
+		read -p "DANGER!!! Will perform a kdump test by crashing the system, proceed? (Y/N): " input
```
prudo1, line 1767: Two things:
1) I prefer giving the different options as lower case letters and indicate the default as upper case letter. I.e. 's;Y/N;y/N;'
2) I understand that you want to give users a chance to abort in case they have second thoughts. But that only makes sense when users trigger the test manually. When they use it in a test script this will cause issues. Does it make sense to add a --force option to disable this question?

mkdumprd
```diff
@@ -60,9 +60,12 @@ add_dracut_sshkey()
...
 to_mount()
 {
-	local _target=$1 _fstype=$2 _options=$3 _sed_cmd _new_mntpoint _pdev
+	local _target=$1 _fstype=$2 _options=$3 _new_mntpoint=$4
+	local _sed_cmd _pdev
 
-	_new_mntpoint=$(get_kdump_mntpoint_from_target "$_target")
+	if [[ -z "$_new_mntpoint" ]]; then
+		_new_mntpoint=$(get_kdump_mntpoint_from_target "$_target")
+	fi
```
prudo1, line 68:
```
_new_mntpoint="${_new_mntpoint:-$(get_kdump_mntpoint_from_target "$_target")}"
```
is identical but shorter and better matches what is done for the other options right below.

mkdumprd
```diff
@@ -414,4 +417,24 @@ if ! is_fadump_capable; then
...
+# sysroot.mount unit will be generated based on 'root=' kernel cmdline, and
+# its mount r/w option will be taken from cmdline too. If we are mounting
+# the same device on both /sysroot and /vmcorestatus, we need to take the same
+# r/w option, otherwise the systemd mounting will fail. So we copy sysroot's
+# r/w option here, and remount it as writeable during kdump capture service
+# when needed.
+if is_fadump_capable && \
+   [[ $(get_root_fs_device) == "$status_target" ]] && \
+   [[ $(cat /proc/cmdline) =~ $ro_regex ]]; then
+        add_mount "$status_target" "" "ro " "$new_mntpoint"
```
prudo1, line 435: Does adding the `"ro"` here have any impact? In `to_mount` the `ro` should be converted back to `rw` anyway.

liutgnu:
Hi Philipp,

Thanks a lot for your careful review!

> Hi Tao, here are some comments from my first round. I still need to think about a few things, e.g. I don't know if it really makes sense to store the status file on the rootfs. My first thought was that it would be better to store it on the target, next to the dumps instead. IMHO, that would make tracking the status easier, when you have a big fleet dumping to a remote host, e.g. via ssh. But I'm not entirely sure if we can assume that the target is always mounted/accessible. Especially when dumping to nfs.

Yes, saving the status file onto the target was my first option, but I gave it up due to the 2 reasons: 1) It is difficult to identify 'who created the vmcore/status file'. Say if we have a nfs/ssh server, several machines will dump vmcore/status file to it. Since we need to verify if kdump test is successful for one specific machine, we need to identify the vmcore/status file is created by it, no by others. Frankly there is no stable way to do that, the ip+date string isn't a good method, e.g the following case: physical machine_a hosted vm1 and vm2 by its NAT network, physical machine_b hosted vm3 and vm4 by its NAT network, and physical machine_c is for vmcore hosting. So vm1 and vm3 can have the same ip address as 10.0.2.15. How can we know if one vmcore/status file belongs to vm1 or vm3? My previous solution is to create a special string(a timestamp) in kernel before trigger kernel crash, and check the vmcore-dmesg.txt file if the string exists, to identify its belonging.

2) Though [1] is doable, but it need to access the vmcore server frenqently(each checking of kdump test success or not), QE/customers don't want us to access the vmcore server too much, unless we are dumping vmcore to it. In addition, accessing a remote server is slow, especially when there are hundreds of vmcores, which means we need to find the special string within hundreds vmcore-dmesg.txt files. It is slow, and nfs-server may not mounted as well...

So saving the status file locally is fast and have no problem to determine its belonging than saving remotely.

kdump-lib-initramfs.sh
```diff
@@ -171,3 +172,34 @@ kdump_get_ip_route_field()
...
 	echo "$1" | sed -n -e "s/^.*\<$2\>\s\+\(\S\+\).*$/\1/p"
 }
+
+# $1: success/fail/clear
+# $2: status_file
+set_vmcore_creation_status()
+{
+	_status=$1
+	_status_file=$2
+	_dir=${_status_file%%${_status_file##*/}}
```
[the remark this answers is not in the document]
liutgnu: OK agreed, we need to install dirname cmd into initramfs first though...

kdumpctl
```diff
@@ -1654,6 +1657,31 @@ _should_reset_crashkernel() {
 	[[ $(kdump_get_conf_val auto_reset_crashkernel) != no ]] && systemctl is-enabled kdump &> /dev/null
 }
 
+check_vmcore_creation_status()
+{
+	local _record _status _timestamp _status_date
+
+	if [[ -z "$VMCORE_CREATION_NOTIFICATION" ]]; then
+		return
+	fi
```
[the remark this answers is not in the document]
liutgnu: OK agreed, will get it improved in v7.

kdumpctl
```diff
@@ -1654,6 +1657,31 @@ _should_reset_crashkernel() {
...
+	if [[ ! -s $VMCORE_CREATION_STATUS ]]; then
+		dwarn "Notice: No vmcore creation test performed!"
+		return
+	fi
+
+	_record=$(cat "$VMCORE_CREATION_STATUS")
+	_status=$(echo $_record | awk '{print $1}')
+	_timestamp=$(echo $_record | awk '{print $2}')
+	_status_date=$(echo "@"$_timestamp | xargs date -d 2> /dev/null)
+
```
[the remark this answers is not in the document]
liutgnu: Thanks, your suggestion is much better!

kdumpctl
```diff
@@ -1729,8 +1758,27 @@ main()
 			reset_crashkernel_for_installed_kernel "$2"
 		fi
 		;;
+	test)
+		is_kernel_loaded "$DEFAULT_DUMP_MODE"
+		if [[ "$?" -ne 0 ]]; then
```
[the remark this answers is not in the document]
liutgnu: Agreed

kdumpctl
```diff
@@ -1729,8 +1758,27 @@ main()
...
+			derror "Kdump should be operational before test."
```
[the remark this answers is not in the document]
liutgnu: Agreed

kdumpctl
liutgnu: OK agreed, I have the same thought when coding...

kdumpctl
liutgnu: Yes, I agree with your "--force" option, it is better!

mkdumprd
liutgnu: Agreed

mkdumprd
liutgnu: No it won't, you see it is "ro ", not "ro", there is one extra space within, so it won't be converted to "rw" later in to_mount(). Personally I don't like the "hacking" here, but didn't think of a better approach...

kdump-lib-initramfs.sh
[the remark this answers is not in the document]
prudo1: right... but I believe it's worth it...

kdumpctl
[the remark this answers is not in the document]
prudo1: Thanks

kdumpctl
prudo1: Thanks

mkdumprd
prudo1: Arghh, ok I missed the extra space... That is really hacky...

Can't we use an different approach instead? If I understand the comment above correctly this hack is only needed when we dump to the rootfs. So we could only add the new mount point when it's _not_ a dump to rootfs and then check in dracut-kdump.sh whether the new mount point exists or not. If it exists we write the status there otherwise it must be a dump to rootfs and we can simply add the status to the target.
This should work, shouldn't it?

prudo1:
> Hi Philipp,
> 
> Thanks a lot for your careful review!
> 
> > Hi Tao, here are some comments from my first round. I still need to think about a few things, e.g. I don't know if it really makes sense to store the status file on the rootfs. My first thought was that it would be better to store it on the target, next to the dumps instead. IMHO, that would make tracking the status easier, when you have a big fleet dumping to a remote host, e.g. via ssh. But I'm not entirely sure if we can assume that the target is always mounted/accessible. Especially when dumping to nfs.
> 
> Yes, saving the status file onto the target was my first option, but I gave it up due to the 2 reasons: 1) It is difficult to identify 'who created the vmcore/status file'. Say if we have a nfs/ssh server, several machines will dump vmcore/status file to it. Since we need to verify if kdump test is successful for one specific machine, we need to identify the vmcore/status file is created by it, no by others. Frankly there is no stable way to do that, the ip+date string isn't a good method, e.g the following case: physical machine_a hosted vm1 and vm2 by its NAT network, physical machine_b hosted vm3 and vm4 by its NAT network, and physical machine_c is for vmcore hosting. So vm1 and vm3 can have the same ip address as 10.0.2.15. How can we know if one vmcore/status file belongs to vm1 or vm3? My previous solution is to create a special string(a timestamp) in kernel before trigger kernel crash, and check the vmcore-dmesg.txt file if the string exists, to identify its belonging.

Right, the ip+date is a pretty poor method to identify the system that dumped. My idea was to replace the ip with the machine-id. The machine-id is determined at the first boot after installation and then stays stable for ever. At least for "normal" installations. I'm not a 100% sure how it behaves for container images where the rootfs is ro. If I understand the man page correctly the machine-id is then determined every time on boot. So it won't be stable. But in that scenario the ip probably wouldn't be stable as well.
So we could switch from ip+date to machine-id+date for the dumps and machine-id.status for the status files. Then we also don't need to check all the vmcore-dmesg.txt.

What do you think?

> 
>     2. Though [1] is doable, but it need to access the vmcore server frenqently(each checking of kdump test success or not), QE/customers don't want us to access the vmcore server too much, unless we are dumping vmcore to it. In addition, accessing a remote server is slow, especially when there are hundreds of vmcores, which means we need to find the special string within hundreds vmcore-dmesg.txt files. It is slow, and nfs-server may not mounted as well...

True, but at least for ssh targets we already connect to them every time we start to check whether the network is up and running. So in that case there won't be too many additional connections. But then there is `kdumpctl status` where we don't know whether we can reach the target. And of course nfs targets...

> So saving the status file locally is fast and have no problem to determine its belonging than saving remotely.

True, but it also means that we need to mount a disk locally for cases where we didn't need it before. Which will require additional memory and thus increases the chance that the kdump kernel runs oom...

liutgnu:
> > Hi Philipp,
> > Thanks a lot for your careful review!
> > > Hi Tao, here are some comments from my first round. I still need to think about a few things, e.g. I don't know if it really makes sense to store the status file on the rootfs. My first thought was that it would be better to store it on the target, next to the dumps instead. IMHO, that would make tracking the status easier, when you have a big fleet dumping to a remote host, e.g. via ssh. But I'm not entirely sure if we can assume that the target is always mounted/accessible. Especially when dumping to nfs.
> > 
> > 
> > Yes, saving the status file onto the target was my first option, but I gave it up due to the 2 reasons: 1) It is difficult to identify 'who created the vmcore/status file'. Say if we have a nfs/ssh server, several machines will dump vmcore/status file to it. Since we need to verify if kdump test is successful for one specific machine, we need to identify the vmcore/status file is created by it, no by others. Frankly there is no stable way to do that, the ip+date string isn't a good method, e.g the following case: physical machine_a hosted vm1 and vm2 by its NAT network, physical machine_b hosted vm3 and vm4 by its NAT network, and physical machine_c is for vmcore hosting. So vm1 and vm3 can have the same ip address as 10.0.2.15. How can we know if one vmcore/status file belongs to vm1 or vm3? My previous solution is to create a special string(a timestamp) in kernel before trigger kernel crash, and check the vmcore-dmesg.txt file if the string exists, to identify its belonging.
> 
> Right, the ip+date is a pretty poor method to identify the system that dumped. My idea was to replace the ip with the machine-id. The machine-id is determined at the first boot after installation and then stays stable for ever. At least for "normal" installations. I'm not a 100% sure how it behaves for container images where the rootfs is ro. If I understand the man page correctly the machine-id is then determined every time on boot. So it won't be stable. But in that scenario the ip probably wouldn't be stable as well. So we could switch from ip+date to machine-id+date for the dumps and machine-id.status for the status files. Then we also don't need to check all the vmcore-dmesg.txt.
> 
> What do you think?

I guess we cannot do that, the machine-id should be encrpted and shouldn't be exposed to outeside. See the "man machine-id", as I quote some of it:

This ID uniquely identifies the host. It should be considered "confidential", and must not be exposed in untrusted environments, in particular on the network...

If we do using some hash/encrypt approach before using machine-id, I guess it will introduce extra complexity for kdump function.

> 
> > ```
> > 2. Though [1] is doable, but it need to access the vmcore server frenqently(each checking of kdump test success or not), QE/customers don't want us to access the vmcore server too much, unless we are dumping vmcore to it. In addition, accessing a remote server is slow, especially when there are hundreds of vmcores, which means we need to find the special string within hundreds vmcore-dmesg.txt files. It is slow, and nfs-server may not mounted as well...
> > ```
> 
> True, but at least for ssh targets we already connect to them every time we start to check whether the network is up and running. So in that case there won't be too many additional connections. But then there is `kdumpctl status` where we don't know whether we can reach the target. And of course nfs targets...

Yeah, since CEE will use "kdumpctl test" often, and I guess nfs targets are their priority. So we need to come up with a universal approach, not only for local/ssh, but also nfs target...
> 
> > So saving the status file locally is fast and have no problem to determine its belonging than saving remotely.
> 
> True, but it also means that we need to mount a disk locally for cases where we didn't need it before. Which will require additional memory and thus increases the chance that the kdump kernel runs oom...

Yes agreed, but I guess it is OK to increase the creashkernel size a little bit in order to support this.

mkdumprd
liutgnu: OK, thanks for the suggestion, I will draft the code and see it can work as expected.

kdump-lib-initramfs.sh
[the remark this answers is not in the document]
liutgnu: OK agreed, will add it in next patch.

prudo1:
> > > Hi Philipp,
> > > Thanks a lot for your careful review!
> > > > Hi Tao, here are some comments from my first round. I still need to think about a few things, e.g. I don't know if it really makes sense to store the status file on the rootfs. My first thought was that it would be better to store it on the target, next to the dumps instead. IMHO, that would make tracking the status easier, when you have a big fleet dumping to a remote host, e.g. via ssh. But I'm not entirely sure if we can assume that the target is always mounted/accessible. Especially when dumping to nfs.
> > > 
> > > 
> > > Yes, saving the status file onto the target was my first option, but I gave it up due to the 2 reasons: 1) It is difficult to identify 'who created the vmcore/status file'. Say if we have a nfs/ssh server, several machines will dump vmcore/status file to it. Since we need to verify if kdump test is successful for one specific machine, we need to identify the vmcore/status file is created by it, no by others. Frankly there is no stable way to do that, the ip+date string isn't a good method, e.g the following case: physical machine_a hosted vm1 and vm2 by its NAT network, physical machine_b hosted vm3 and vm4 by its NAT network, and physical machine_c is for vmcore hosting. So vm1 and vm3 can have the same ip address as 10.0.2.15. How can we know if one vmcore/status file belongs to vm1 or vm3? My previous solution is to create a special string(a timestamp) in kernel before trigger kernel crash, and check the vmcore-dmesg.txt file if the string exists, to identify its belonging.
> > 
> > 
> > Right, the ip+date is a pretty poor method to identify the system that dumped. My idea was to replace the ip with the machine-id. The machine-id is determined at the first boot after installation and then stays stable for ever. At least for "normal" installations. I'm not a 100% sure how it behaves for container images where the rootfs is ro. If I understand the man page correctly the machine-id is then determined every time on boot. So it won't be stable. But in that scenario the ip probably wouldn't be stable as well. So we could switch from ip+date to machine-id+date for the dumps and machine-id.status for the status files. Then we also don't need to check all the vmcore-dmesg.txt.
> > What do you think?
> 
> I guess we cannot do that, the machine-id should be encrpted and shouldn't be exposed to outeside. See the "man machine-id", as I quote some of it:
> 
> This ID uniquely identifies the host. It should be considered "confidential", and must not be exposed in untrusted environments, in particular on the network...

Yeah, I read that as well. But I believe that for our use case we can assume that the dump target can be trusted even when it's a remote target. Reason is that, if we cannot trust the dump target we cannot write the dump there as well. In the end a kernel dump will always contain a lot of confidential information. So if we write a dump to an untrusted target the machine-id is one of the smaller problems we have.

> If we do using some hash/encrypt approach before using machine-id, I guess it will introduce extra complexity for kdump function.

Fully agree, but as said above IMHO that shouldn't be necessary.

> > > ```
> > > 2. Though [1] is doable, but it need to access the vmcore server frenqently(each checking of kdump test success or not), QE/customers don't want us to access the vmcore server too much, unless we are dumping vmcore to it. In addition, accessing a remote server is slow, especially when there are hundreds of vmcores, which means we need to find the special string within hundreds vmcore-dmesg.txt files. It is slow, and nfs-server may not mounted as well...
> > > ```
> > 
> > 
> > True, but at least for ssh targets we already connect to them every time we start to check whether the network is up and running. So in that case there won't be too many additional connections. But then there is `kdumpctl status` where we don't know whether we can reach the target. And of course nfs targets...
> 
> Yeah, since CEE will use "kdumpctl test" often, and I guess nfs targets are their priority. So we need to come up with a universal approach, not only for local/ssh, but also nfs target...

True...

> > > So saving the status file locally is fast and have no problem to determine its belonging than saving remotely.
> > 
> > 
> > True, but it also means that we need to mount a disk locally for cases where we didn't need it before. Which will require additional memory and thus increases the chance that the kdump kernel runs oom...
> 
> Yes agreed, but I guess it is OK to increase the creashkernel size a little bit in order to support this.

Yes, I guess so...

liutgnu:
Now the v7 is posted

liutgnu:
> > > > Hi Philipp,
> > > > Thanks a lot for your careful review!
> > > > > Hi Tao, here are some comments from my first round. I still need to think about a few things, e.g. I don't know if it really makes sense to store the status file on the rootfs. My first thought was that it would be better to store it on the target, next to the dumps instead. IMHO, that would make tracking the status easier, when you have a big fleet dumping to a remote host, e.g. via ssh. But I'm not entirely sure if we can assume that the target is always mounted/accessible. Especially when dumping to nfs.
> > > > 
> > > > 
> > > > Yes, saving the status file onto the target was my first option, but I gave it up due to the 2 reasons: 1) It is difficult to identify 'who created the vmcore/status file'. Say if we have a nfs/ssh server, several machines will dump vmcore/status file to it. Since we need to verify if kdump test is successful for one specific machine, we need to identify the vmcore/status file is created by it, no by others. Frankly there is no stable way to do that, the ip+date string isn't a good method, e.g the following case: physical machine_a hosted vm1 and vm2 by its NAT network, physical machine_b hosted vm3 and vm4 by its NAT network, and physical machine_c is for vmcore hosting. So vm1 and vm3 can have the same ip address as 10.0.2.15. How can we know if one vmcore/status file belongs to vm1 or vm3? My previous solution is to create a special string(a timestamp) in kernel before trigger kernel crash, and check the vmcore-dmesg.txt file if the string exists, to identify its belonging.
> > > 
> > > 
> > > Right, the ip+date is a pretty poor method to identify the system that dumped. My idea was to replace the ip with the machine-id. The machine-id is determined at the first boot after installation and then stays stable for ever. At least for "normal" installations. I'm not a 100% sure how it behaves for container images where the rootfs is ro. If I understand the man page correctly the machine-id is then determined every time on boot. So it won't be stable. But in that scenario the ip probably wouldn't be stable as well. So we could switch from ip+date to machine-id+date for the dumps and machine-id.status for the status files. Then we also don't need to check all the vmcore-dmesg.txt.
> > > What do you think?
> > 
> > 
> > I guess we cannot do that, the machine-id should be encrpted and shouldn't be exposed to outeside. See the "man machine-id", as I quote some of it:
> > This ID uniquely identifies the host. It should be considered "confidential", and must not be exposed in untrusted environments, in particular on the network...
> 
> Yeah, I read that as well. But I believe that for our use case we can assume that the dump target can be trusted even when it's a remote target. Reason is that, if we cannot trust the dump target we cannot write the dump there as well. In the end a kernel dump will always contain a lot of confidential information. So if we write a dump to an untrusted target the machine-id is one of the smaller problems we have.

OK, agreed. However one thing I noticed from my fedora:37 container, the systemd package is not installed:
[root@3c6570f17fba system]# rpm -qa|grep systemd
systemd-libs-251.10-588.fc37.x86_64

So there is no systemd-machine-id-commit.service to generate the machine-id, instead we got and empty file: 
[root@3c6570f17fba system]# rpm -qf /etc/machine-id
file /etc/machine-id is not owned by any package
[root@3c6570f17fba system]# ls -ail /etc/machine-id
3046437 -rw-r--r--. 1 root root 0 Feb  5  2023 /etc/machine-id 

I guess machine-id is not a stable approach for containers than ip+date...

> 
> > If we do using some hash/encrypt approach before using machine-id, I guess it will introduce extra complexity for kdump function.
> 
> Fully agree, but as said above IMHO that shouldn't be necessary.
> 
> > > > ```
> > > > 2. Though [1] is doable, but it need to access the vmcore server frenqently(each checking of kdump test success or not), QE/customers don't want us to access the vmcore server too much, unless we are dumping vmcore to it. In addition, accessing a remote server is slow, especially when there are hundreds of vmcores, which means we need to find the special string within hundreds vmcore-dmesg.txt files. It is slow, and nfs-server may not mounted as well...
> > > > ```
> > > 
> > > 
> > > True, but at least for ssh targets we already connect to them every time we start to check whether the network is up and running. So in that case there won't be too many additional connections. But then there is `kdumpctl status` where we don't know whether we can reach the target. And of course nfs targets...
> > 
> > 
> > Yeah, since CEE will use "kdumpctl test" often, and I guess nfs targets are their priority. So we need to come up with a universal approach, not only for local/ssh, but also nfs target...
> 
> True...
> 
> > > > So saving the status file locally is fast and have no problem to determine its belonging than saving remotely.
> > > 
> > > 
> > > True, but it also means that we need to mount a disk locally for cases where we didn't need it before. Which will require additional memory and thus increases the chance that the kdump kernel runs oom...
> > 
> > 
> > Yes agreed, but I guess it is OK to increase the creashkernel size a little bit in order to support this.
> 
> Yes, I guess so...

liutgnu:
Hi @prudo1 , kindly ping. Could you please review the patch? If you have no time, then we can ask other's help. Thanks!

dracut-kdump.sh
```diff
@@ -312,11 +312,24 @@ do_final_action()
 
 do_dump()
 {
+	local _vmcore_creation_status
+
```
prudo1, line 316: `kdump.sh` runs in the initrd and needs to be POSIX compliant. You cannot use `local` here.

kdumpctl
```diff
@@ -1654,6 +1657,65 @@ _should_reset_crashkernel() {
...
+
+	_dir=$(dirname "$VMCORE_CREATION_STATUS")
+	if ! [[ -d "$_dir" ]]; then
+		derror "Vmcore status dir $_dir not exist."
+		exit 1
+	fi
+
+	lsblk $(get_mount_info SOURCE target "$_dir") > /dev/null
+
+	if [[ $? -ne 0 ]]; then
```
prudo1, line 1698: You can check the return value directly with
```
if ! lsblk $(get_mount_info SOURCE target "$_dir") > /dev/null; then
...
```

kdumpctl
```diff
@@ -1729,8 +1792,12 @@ main()
 			reset_crashkernel_for_installed_kernel "$2"
 		fi
 		;;
+	test)
+		shift
+		kdump_test "$@"
+		;;
 	*)
-		dinfo $"Usage: $0 {estimate|start|stop|status|restart|reload|rebuild|reset-crashkernel|propagate|showmem}"
+		dinfo $"Usage: $0 {estimate|start|stop|status|restart|reload|rebuild|reset-crashkernel|propagate|showmem|test [--force]}"
```
prudo1, line 1800: I would drop the `--force` here as the other sub-commands don't specify their options here as well.
That should be done in `kdumpctl.8` where you should add a section for the new command anyway ;-)

mkdumprd
```diff
@@ -414,4 +415,17 @@ if ! is_fadump_capable; then
...
 fi
 
+status_target=$(get_target_from_path $(dirname "$VMCORE_CREATION_STATUS"))
+
+if [[ $(get_root_fs_device) != "$status_target" ]]; then
+	new_mntpoint=$(echo /vmcorestatus/$(get_mntpoint_from_target "$status_target") \
+		| tr -s "/")
+	add_mount "$status_target" "" "" "$new_mntpoint"
+else
+	if ! is_fadump_capable && \
```
prudo1, line 425:
```
elif ! is_fadump_capable && \
...
```

prudo1 (changes requested):
Hi @liutgnu,
this PR is almost perfect. Just a few very small nits. The biggest "problems" I see is that with #33 merged you need to rebase to the latest main, as there will be some conflicts. In addition you need to add the new sub-command to `kdumpctl.8` man page. With that fixed the PR can be merged.
Thanks!

liutgnu:
> Hi @liutgnu, this PR is almost perfect. Just a few very small nits. The biggest "problems" I see is that with #33 merged you need to rebase to the latest main, as there will be some conflicts. In addition you need to add the new sub-command to `kdumpctl.8` man page. With that fixed the PR can be merged. Thanks!

Hi @prudo1 , thanks a lot for your review. I will get those improved in v8

dracut-kdump.sh
liutgnu: Thanks, agreed.

kdumpctl
liutgnu: Thanks, much better!

mkdumprd
liutgnu: Agreed

kdumpctl
liutgnu: Thanks for the suggestion, agreed, will update the man entry as well.

liutgnu:
> > Hi @liutgnu, this PR is almost perfect. Just a few very small nits. The biggest "problems" I see is that with #33 merged you need to rebase to the latest main, as there will be some conflicts. In addition you need to add the new sub-command to `kdumpctl.8` man page. With that fixed the PR can be merged. Thanks!
> 
> Hi @prudo1 , thanks a lot for your review. I will get those improved in v8

Hi @prudo1 , please see the v8 I just posted. Thanks in advance!

prudo1 (approved):
@liutgnu looks good now. Thanks!

daveyoung (approved):

PR #26, opened 2024-07-12
Temporary disable PrivateTmp=yes

licliu:
Using Privatetmp=yes will cause mountinfo changes, which will cause kdump started via systemd to fail in scenarios using bind mount such as coreos. Temporarily disable it until the bind mount problem is resolved.

Fixes: ea00b7db("kdumpctl: Move temp file in get_kernel_size to global temp dir")


prudo1 (approved):
Hi @licliu, looks good to me. Thanks!

coiby (approved):

PR #28, opened 2024-07-23
Fix compatibility with bootc systems

licliu:


kdump-lib.sh
liutgnu, line 1: I think maybe the line modified is incorrect here, the code change in the context doesn't make scence here, can you verify it? The modified line should be:

```
	_fsroot=${_src#${_src_nofsroot}[}
	_fsroot=${_fsroot%]}
-	_mnt=$(get_mount_info TARGET source "$_src_nofsroot" -f)
+	_mnt=$(get_mntpoint_from_target "$_src_nofsroot")
```

liutgnu:
For the v2, ack

kdump-lib-initramfs.sh
```diff
@@ -102,8 +102,15 @@ get_fs_type_from_target()
...
-	get_mount_info TARGET source "$1" -f
+	local SOURCE TARGET
+	findmnt -k --pairs -o SOURCE,TARGET "$1" | while read line; do
+		eval "$line"
+		# omit sources that are bind mounts i.e. they contain a [/path/to/subpath].
+		if [[ ! "$SOURCE" =~ \[ ]]; then
+			echo $TARGET
+			break
+		fi
+	done
```
prudo1, line 113: I really like to avoid 'eval' here. Especially as it's not necessary.
For me the following worked
```
while read _source _target; do
    ...
done < <(findmnt -k -n -o SOURCE,TARGET "$1")
```
Other than that the series looks fine to me.

prudo1 (changes requested):

kdump-lib-initramfs.sh
cgwalters: That would fail if the source or target contained whitespace, which is pretty unlikely of course...but, still.

(OTOH, the real fix is to not write any nontrivial shell script at all...)

kdump-lib.sh
cgwalters: I may be missing something...what's the difference between the diff in your comment and the proposed diff?

liutgnu: Never mind, lichen made a mistake in v1 about this hunk and fixed in v2...

kdump-lib-initramfs.sh
licliu: Hi @prudo1 , I think we can merge this first since it works well now. And we can make changes later to avoid code conflicts if we touch this code in the future.

licliu:
Hi @prudo1, I changed the code to avoid 'eval'. And shellcheck doesn't like `done < <(program)` so I still use pipe.

dracut-module-setup.sh
```diff
@@ -1041,6 +1041,9 @@ install() {
     # Also redirect dracut-emergency to kdump error handler
     ln_r "$systemdsystemunitdir/emergency.service" "$systemdsystemunitdir/dracut-emergency.service"
 
+    # Disable ostree as we only need the physical root
+    systemctl -q --root "$initdir" mask ostree-prepare-root.service
```
travier, line 1045: If you do this, this means that the crash dump file will be written in the "real root" partition (`/sysroot`) and not in the actual location set in the configuration file.

cgwalters: I did test this. I think the reason it works is because the earlier code is resolving through the bind mount to the real sysroot target.

kdump-lib-initramfs.sh
```diff
@@ -102,8 +102,20 @@ get_fs_type_from_target()
...
-	get_mount_info TARGET source "$1" -f
+	local _source _target
+	# omit sources that are bind mounts i.e. they contain a [/path/to/subpath].
+	findmnt -k --pairs -o SOURCE,TARGET "$1" | while read -r line; do
+		_source="${line#SOURCE=\"}"
+		_source="${_source%%\" TARGET=*}"
+
+		_target="${line#*TARGET=\"}"
+		_target="${_target%%\"}"
+
```
prudo1, line 113: Hi Lichen,
a few nits here.
1.  shellcheck complains about the use of `local` as it's not POSIX compliant
2. you can read multiple variables with `read`, i.e. you can drop `line`
3. when you drop `--pairs` `findmnt` should omit the `SOURCE=` and `TARGET=` prefix as well as the quotation (please double check)

putting it all together this should simplify the code to a single line
```
findmnt -k -o SOURCE,TARGET "$1" | while read -r _source _target; do
```
Other than that the series looks fine

prudo1: Oups, missed the comment from @cgwalters above.  He's right, that 3. would break if either `_source` or `_target` contains a space. So probably better to keep option `--pairs`.

licliu: Hi @prudo1 ,thanks for reviewing. I removed the `local _source _target` and use `read -r _source _target` now.
I keep `--pairs` becasue in my test, I need to handle cases when `_source` or `_target` contains a space.

kdump-lib-initramfs.sh
coiby, line 1: Unfortunately, this patch has several issues,
1. it will cause the same regression https://issues.redhat.com/browse/RHEL-56832 as the previous [RHEL-only](https://gitlab.com/redhat/centos-stream/rpms/kexec-tools/-/commit/099aead590608cd9edb49acf73873b03b88576b6) will `expr "X$_source" : 'X.*\['` will also match an IPv6  target specified in kdump.conf e.g. `nfs [2203...]:/nfs_server`
2. The regex also matches brtfs subvol so it will return empty value for btrfs target
```
[root@localhost ~]# findmnt -k --pairs -o SOURCE,TARGET /dev/vda4
SOURCE="/dev/vda4[/root]" TARGET="/"
SOURCE="/dev/vda4[/home]" TARGET="/home"
SOURCE="/dev/vda4[/var]" TARGET="/var"
SOURCE="/dev/vda4[/var]" TARGET="/tt2"
```

I think a better way to to fix `get_bind_mount_source` is to check `/proc/self/mountinfo`,
```
[root@localhost ~]# mkdir /tt2
[root@localhost ~]# mount --bind /var /tt2

[root@localhost ~]# mkdir "/tt2 space"
[root@localhost ~]# mount --bind /var "/tt2 space"

[root@localhost ~]# grep tt2 /proc/self/mountinfo 
192 77 0:39 /var /tt2 rw,relatime shared:118 - btrfs /dev/vda4 rw,seclabel,compress=zstd:1,discard=async,space_cache=v2,subvolid=258,subvol=/var
488 77 0:39 /var /tt2\040space rw,relatime shared:118 - btrfs /dev/vda4 rw,seclabel,compress=zstd:1,discard=async,space_cache=v2,subvolid=258,subvol=/var

```
Btw, The `get_mntpoint_from_target` names suggests it returns a mount point based on a target. But what it does is return a mount point based on a source. It should be renamed in order to not confuse readers.

coiby (changes requested):

kdump-lib-initramfs.sh
coiby: To fix the regression first, I've created [a centos-stream kdump-utils MR](https://gitlab.com/redhat/centos-stream/rpms/kexec-tools/-/merge_requests/241). But we need to fix the btrfs subvol issue. 

Btw, there may be another problem with this PR. In this PR  `get_mntpoint_from_target` no longer call `get_mount_info` which will fallback to find mount info in `/etc/fstab` and only call `findmnt` with `-k` argument. So this PR's `get_mntpoint_from_target` will not be able to make use of `/etc/fstab` and may cause problem for specifying a not-mounted target.

licliu: Hi @coiby , maybe we can use FSTYPE to  handle this case.
For btrfs, it is in the Unsupported Dump targets and we have a [fedora bug](https://bugzilla.redhat.com/show_bug.cgi?id=2270423). 
Maybe we can take this opportunity to refactor our code to handle different file systems.

Issue #11, opened 2024-06-11
kdump should auto-add required modules if they are missing from host initrd


jbtrystram:
In Fedora CoreOS, we exclude nfs from the initramfs as booting from NFS is not supported : https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/dracut.conf.d/coreos-omits.conf#L7

However, dumping with kdump to a target destination should work obviously. Currently, this requires the users to add 
`extra_modules nfs` to `/etc/kdump.conf`. 

Could the module be added automatically when the target is a NFS destination ? 

ref: https://github.com/coreos/fedora-coreos-tracker/issues/1729

daveyoung:
Hi, Coiby filed an RH internal jira issue to dracut,  dracut team suggest to use --force-add to override the config file
dracut --force-add nfs
Can you try see if this works I think we can do it in kdump code.   Please refer to https://issues.redhat.com/browse/RHEL-26114

jbtrystram:
Hello,

I did the following testing today, i was not able to set-up a NFS server for testing so I can only attest what
modules are loaded according to config options. 

on FCOS, with the default config the NFS module is not loaded in kdump initramfs : 
```
kdump:/# modinfo nfs
modinfo: ERROR: Module nfs not found.
```

With the following `kdump.conf` config : 
```
          path /var/crash
          extra-modules nfs
```
I can see the module loaded:
```
kdump:/# modinfo nfs
filename:       /lib/modules/6.8.11-300.fc40.x86_64/kernel/fs/nfs/nfs.ko.xz
license:        GPL
author:         Olaf Kirch <okir@monad.swb.de>
alias:          nfs4
alias:          fs-nfs4
alias:          fs-nfs
rhelversion:    9.99
depends:        sunrpc,netfs,lockd
retpoline:      Y
intree:         Y
name:           nfs
vermagic:       6.8.11-300.fc40.x86_64 SMP preempt mod_unload 
sig_id:         PKCS#7
signer:         Fedora kernel signing key
sig_key:        12:8F:6B:9F:1E:CC:CB:0D:BA:F0:10:12:70:1C:A6:DD:A1:09:9A:11
sig_hashalgo:   sha256
signature:      CC:3........
```

However, using `dracut-args` like this does not works: 
```
          path /var/crash
         dracut-args --force-add nfs
```

During the kdump initramfs build dracut fails : 
```
[...]

PR #29, opened 2024-07-30
mkdumprd: Skip global config setting dracutmodules=

cgwalters:
The Fedora/RHEL bootc base images have dracut drop-ins which set `dracutmodules+=`.

However that seems override explicit module inclusion on the command line, which is a dracut bug:
https://github.com/rhkdump/kdump-utils/issues/11

Work around this by making our own copy of the global config, and omitting config drop-ins which trigger this behavior.

I think longer term, this project should probably own its own global dracut config, actually. But that's a much larger set of work.

### DIFF
--- a/mkdumprd
+++ b/mkdumprd
@@ -319,9 +319,40 @@ check_crypt()
 	done
 }
 
+temp_dracut_confdir=
+# This is a workaround for dracut argument inheritance not doing
+# what one would expect. xref https://issues.redhat.com/browse/RHEL-49590
+# and https://github.com/rhkdump/kdump-utils/issues/11
+maybe_synthesize_tmp_confdir()
+{
+	local confs
+	confs=$(shopt -s nullglob; echo /usr/lib/dracut/dracut.conf.d/*.conf)
+	if test -z "${confs}"; then
+		return
+	fi
+	temp_dracut_confdir=$(mktemp -d -t kdump.XXXXXX)
+	for c in ${confs}; do
+		if grep -qEe '^(omit_|)dracutmodules.*=' "${c}"; then
+			echo "Skipping system config which overrides dracutmodules: ${c}"
+			continue
+		fi
+		echo "Using config: ${c}"
+		cp -p $c ${temp_dracut_confdir}/
+	done
+	echo "Synthesized temp configdir: ${temp_dracut_confdir}"
+	add_dracut_arg --confdir "${temp_dracut_confdir}"
+	# and now ensure dracut can't see our overrides. This relies on
+	# kdump.service being in a new mount namespace.
+	mount --bind -o ro /usr/share/empty /usr/lib/dracut/dracut.conf.d
+}
+
 if ! check_crypt; then
 	dwarn "Warning: Encrypted device is in dump path, which is not recommended, see kexec-kdump-howto.txt for more details."
 fi
+
+# Early on, synthesize our own confdir which drops system config
+# that overrides dracutmodules as that causes bugs.
+maybe_synthesize_tmp_confdir
 
 # firstly get right SSH_KEY_LOCATION
 keyfile=$(kdump_get_conf_val sshkey)


cgwalters:
OK, with this I get the `nfs` module in the generated kdump initramfs. However I haven't yet successfully done a kdump to NFS, still testing/debugging that...

cgwalters:
Tested and this works for me; the problem was I didn't have permissions on my NFS server set up correctly for writes.

mkdumprd
```diff
@@ -319,10 +319,41 @@ check_crypt()
...
+maybe_synthesize_tmp_confdir()
+{
+	local confs
+	confs=$(shopt -s nullglob; echo /usr/lib/dracut/dracut.conf.d/*.conf)
+	if test -z "${confs}"; then
+		return
+	fi
+	temp_dracut_confdir=$(mktemp -d -t kdump.XXXXXX)
+	for c in ${confs}; do
+		if grep -qEe '^dracutmodules.*=' "${c}"; then
```
daveyoung, line 335: Hi, it would be better to skip both dracutmodules and omit_dracutmodules as we have the other issue when omit_dracutmodules includes nfs then there's no way to add it for the time being.

licliu:
This works for my virtual machine, but Aman said it didn't work for his bare-metal test enviornment.

cgwalters:
> This works for my virtual machine, but Aman said it didn't work for his bare-metal test enviornment.

Discussion in https://issues.redhat.com/browse/RHEL-49590 narrowed that down to nfsv3 vs nfsv4; do you want to merge this and look at fixes for nfsv3 as a followup?

daveyoung:
> > This works for my virtual machine, but Aman said it didn't work for his bare-metal test enviornment.
> 
> Discussion in https://issues.redhat.com/browse/RHEL-49590 narrowed that down to nfsv3 vs nfsv4; do you want to merge this and look at fixes for nfsv3 as a followup?

Currently I'm not sure about the how hard the dracut fixes needed, and also if this is urgent.  Ideally I still perfer fix dracut,  I put a comment in RHEL-49590, could you and @pvalena provide inputs?

cgwalters:
I think it's somewhat urgent as kdump working in all these scenarios is part of the RHEL certification suite.

cgwalters:
WDYT of my comment:

> I think longer term, this project should probably own its own global dracut config, actually. But that's a much larger set of work.

licliu:
@cgwalters I created #31 for adding kdump dracut config. But for nfsv3 issue, we still need some change in dracut.

coiby:
I'm curious to ask why does boot base image use `dracutmodules` instead of `add_dracutmodules`?
```
       dracutmodules+=" <dracut modules> "
           Specify a space-separated list of dracut modules to call when building the initramfs. Modules are located in /usr/lib/dracut/modules.d. This option forces dracut to only
           include the specified dracut modules. In most cases the "add_dracutmodules" option is what you want to use.
```

cgwalters:
> I'm curious to ask why does boot base image use dracutmodules instead of add_dracutmodules?

This is a good question. Just so everyone's on the same page: by default, dracut does "module autodetection", asking each module it has (including a bunch of built-in ones) whether or not the relevant binaries are present. So you'll basically normally see things like this:

```
dracut[I]: Module 'busybox' will not be installed, because command 'busybox' could not be found!
dracut[I]: Module 'rngd' will not be installed, because command 'rngd' could not be found!
```

(Yes, we should patch dracut to make this look less like an error; maybe condense all of them at the end to a single line `dracut[I]: Skipping modules: 'busybox' 'rngd' ...`)

If we use `add_dracutmodules`, then dracut still does that auto-detection. We could suppress that with `omit_dracutmodules` but then it becomes messier because if someone wants to derive from our image and actually add that content, they'd need to undo that.

---

All this to say...well, I guess we can just switch to `add_dracutmodules`, I can try sending a patch to dracut to clean up the output.

cgwalters:
Ahh actually https://github.com/dracut-ng/dracut-ng/commit/d73cc24e112c01aa701a96a7b8a58adce78409e7 landed in dracut-ng already.

cgwalters:
https://gitlab.com/fedora/bootc/base-images/-/merge_requests/36

coiby:
> > I'm curious to ask why does boot base image use dracutmodules instead of add_dracutmodules?
> 
> This is a good question. Just so everyone's on the same page: by default, dracut does "module autodetection", asking each module it has (including a bunch of built-in ones) whether or not the relevant binaries are present. So you'll basically normally see things like this:
> 
> ```
> dracut[I]: Module 'busybox' will not be installed, because command 'busybox' could not be found!
> dracut[I]: Module 'rngd' will not be installed, because command 'rngd' could not be found!
> ```
> 
> (Yes, we should patch dracut to make this look less like an error; maybe condense all of them at the end to a single line `dracut[I]: Skipping modules: 'busybox' 'rngd' ...`)
> 
> If we use `add_dracutmodules`, then dracut still does that auto-detection. We could suppress that with `omit_dracutmodules` but then it becomes messier because if someone wants to derive from our image and actually add that content, they'd need to undo that.
> 
> All this to say...well, I guess we can just switch to `add_dracutmodules`, I can try sending a patch to dracut to clean up the output.

Thanks for the clarification! So you use dracutmodules mainly to avoid outputting the seeming errors from detecting missing binaries. 

Since 
https://gitlab.com/fedora/bootc/base-images/-/merge_requests/36 has been merged, there is no need for this PR, am I correct?

pvalena:
@coiby FYI if you really want to be sure the module is included, it's better to use `force_add_dracutmodules` which ignores the `omit_dracutmodules` list.

cgwalters:
FTR I created https://issues.redhat.com/browse/RHEL-56076 to track this...it should happen automatically within ~a week.

daveyoung:
I suppose with Lichen's pr #31, this one is not needed and the RHEL-56076 can be worked separately.  Should this pr be closed?  @cgwalters @coiby @licliu

PR #31, opened 2024-08-06
Add kdump dracut config

licliu:
In some cases, we need to change dracut [omit_]dracutmodules to customize the first kernel's initrd to meet the needs, such as bootc and CoreOS.

In most cases kdump does not use these modules, but the change of [omit]_dracutmodules may break existing functionality. For example, if we set omit_dracutmodules='nfs', the nfs module cannot be added even if we use "dracut_args --force-add 'nfs'".

It is better to maintain its own dracut config for kdump. As the first try, we start by keeping [omit_]dracutmodules empty.

As a side effect, the default confdir "/etc/dracut.conf.d" will not work while generating kdump initrd.

And dracut can only use --confdir once, so "dracut_args --confidr" in /etc/kdump.conf will no longer work. This may be avoided in the future by adding something like "--extra-confdir" to dracut.

### DIFF
--- a/99-kdump.conf
+++ b/99-kdump.conf
@@ -1,0 +1,3 @@
+dracutmodules=''
+add_dracutmodules=' kdumpbase '
+omit_dracutmodules=' rdma plymouth resume ifcfg earlykdump '
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ manpages:
 	install -D -m 644 kdump.conf.5 $(DESTDIR)$(mandir)/man5/kdump.conf.5
 
 install: dracut-modules kdump-conf kdump-sysconfig manpages
-	mkdir -p $(DESTDIR)$(pkglibdir)
+	mkdir -p $(DESTDIR)$(pkglibdir)/dracut.conf.d
 	mkdir -p -m755 $(DESTDIR)$(sysconfdir)/kdump/pre.d
 	mkdir -p -m755 $(DESTDIR)$(sysconfdir)/kdump/post.d
 	mkdir -p -m755 $(DESTDIR)$(localstatedir)/crash
@@ -46,6 +46,7 @@ install: dracut-modules kdump-conf kdump-sysconfig manpages
 	install -D -m 644 kdump.conf $(DESTDIR)$(sysconfdir)
 	install -D -m 644 kdump.sysconfig $(DESTDIR)$(sysconfdir)/sysconfig/kdump
 	install -D -m 755 kdump-lib.sh kdump-lib-initramfs.sh kdump-logger.sh -t $(DESTDIR)$(pkglibdir)
+	install -D -m 644 99-kdump.conf -t $(DESTDIR)$(pkglibdir)/dracut.conf.d
 
 ifeq ($(ARCH), $(filter ppc64le ppc64,$(ARCH)))
 	install -m 755 mkfadumprd $(DESTDIR)$(sbindir)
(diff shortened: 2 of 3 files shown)


licliu:
related #29,#11

licliu:
@daveyoung I created a PR for dracut. 
https://github.com/dracut-ng/dracut-ng/pull/569

daveyoung (changes requested):
Hi Lichen, dracut merged the pr already, can you update this one?

licliu:
@daveyoung I think we need to update this patch after dracut-104 is released, otherwise it won't work because the current version of dracut doesn't recognize the --add-confdir option.

daveyoung:
@licliu  good point,  I thought that this is upstream instead of Fedora we can just assume using upstream latest dracut.  But rethinking about it,  dependencies need to be added in kdump-utils.  For example,  if dracut --help|grep "add-confdir" tell us it is supported then we can use the extra confdir, otherwise still use the old way.    Otherwise it could be Fedora only in spec file to depend on dracut 104 and later.  Thoughts?

BTW,  maybe you  can check with dracut team about the dracut 104 schedule plan.

daveyoung (commented):
Hi Lichen, 
 " If --add-confdir is not
available in your system, kdump will use --confdir, in this case,
the default confdir "/etc/dracut.conf.d" will not work while
generating kdump initrd.",  No, I think that we do not need to use  --confdir here.    The thing can be done like this:

if dracut support --add-confdir, then
   use --add-confdir ;
    modify mkdumprd:
    dracut_args=(--add kdumpbase --quiet --hostonly --hostonly-cmdline --hostonly-i18n --hostonly-mode strict --hostonly-nics '' --aggressive-strip -o "plymouth resume ifcfg earlykdump")
    move the args hardcode above to the 99-kdump.conf.
else
    still keep the args in the mkdumprd code like before.

licliu:
@daveyoung Thanks, I will try that in the next version.

licliu:
This version I pushed was just a rebase with conflicts resolved, and is not ready for review yet. 
GitHub seems unable to change PR status back to draft, sorry for the inconvenience. Please look forward to my next push.

licliu:
Hi @daveyoung, I pushed my latest changes just now. I found that dracutmodules variable cannot be override by dracut command line options, so I still use --confdir here.

daveyoung (commented):
Hi Lichen, it seems this new version still does not cover the previous suggestions.   Also the patch description is not aligned with the code, eg. the omit module list is not empy according to the code.  Anyway there might be some misunderstandings. I will try to ping you offline to explain about my suggestions.  Thanks for reworking on this!

pvalena:
FTR `--force-add nfs` should do what you want since https://github.com/redhat-plumbers/dracut-rhel10/pull/18

licliu:
> FTR `--force-add nfs` should do what you want since [redhat-plumbers/dracut-rhel10#18](https://github.com/redhat-plumbers/dracut-rhel10/pull/18)

Thanks! Good to know that.

daveyoung (approved):

daveyoung:
> FTR `--force-add nfs` should do what you want since [redhat-plumbers/dracut-rhel10#18](https://github.com/redhat-plumbers/dracut-rhel10/pull/18)

Hi Pavel, thanks for the info!  This should be used as a workaround, eg. users can use this in kdump.conf dracut_args to force nfs working.   But it is not good to use this in kdump scripts, because there might be other cases not only for nfs, also may need overwrite other things eg. omit modules etc.   So I think we should still wait for the --add-confdir in dracut.   For it is fine that we suggest people to workaround the issue before dracut backport is ready.

PR #34, opened 2024-09-09
mkdumprd: Omit rdma module

pfliu:
Resolves: https://issues.redhat.com/browse/RHEL-57006

05rdma dracut module from rdma-core is installed unconditionally even if kdump dumps the vmcore to local disk. And those kmod will cost additional 200M memory on x86, which likely triggers OOM.

Since the Infiniband (and in fact none of RDMA devices are supported in kdump), exclude the rdma dracut module explicitly.

### DIFF
--- a/mkdumprd
+++ b/mkdumprd
@@ -37,7 +37,7 @@ dracut_args+=(--hostonly-i18n)
 dracut_args+=(--hostonly-mode strict)
 dracut_args+=(--hostonly-nics '')
 dracut_args+=(--aggressive-strip)
-dracut_args+=(--omit "plymouth resume ifcfg earlykdump")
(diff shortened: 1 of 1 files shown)


pfliu:
in rdma-core/redhat/rdma.modules-setup.sh, it just blindless installs all RDMA potential modules

installkernel() {
        hostonly='' instmods =drivers/infiniband =drivers/net/ethernet/mellanox =drivers/net/ethernet/chelsio =drivers/net/ethernet/cisco =drivers/net/ethernet/emulex =drivers/target
        hostonly='' instmods crc-t10dif crct10dif_common xprtrdma svcrdma
}


And if in future, we support RDMA, we need to install the specific driver just like kdump_install_nic_driver() in 99kdumpbase/module-setup.sh

coiby (approved):

PR #33, opened 2024-08-14
Add suport for erofs in the kdump initrd

prudo1:
This series adds support for erofs images in the kdump initrd. The feature was introduced in dracut with this [PR](https://github.com/dracut-ng/dracut-ng/pull/546) and will be included with dracut 104.

While working on the feature I got a little bit side tracked so this series also includes a few cleanups.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rhkdump/kdump-utils/33)

PR #36, opened 2024-09-24
Spec file fix and cleanups

prudo1:
This series is a follow-up from the changes included in PR #33. In particular I noticed when backporting the patches to RHEL that building the rpm failed as `rpmbuild` was trying to patch the wrong file. That's because the spec file contains `%autospec` without option `-p`. This will cause `patch` to strip all directories from the path, which ultimately means that only files in the root directory can be patched by `rpmbuild`. This is fixed in the 2nd commit.

The 3rd commit is a small cleanup for the initial include and should (hopefully) be obvious.

The 1st and 4th commit arise from the question on how we shall manage the spec file in the future. Especially as the spec file contained in this repo and the one from Fedora Rawhide already diverged. If there is no quick answer to the question I suggest to drop these two commits for now and only include the 2nd and 3rd one.

Note: The [rpm docs](https://docs.pagure.org/fedora-infra.rpmautospec/opting-in.html) strongly suggests that the 4th commit should only be included right before a version bump.


coiby (approved):
The PR LGTM, thanks!

PR #38, opened 2024-10-02
kdumpctl: deprecate --reboot for reset-creashkernel

prudo1:
The --reboot option for reset-crashkernel causes quite some confusion. Especially, there are different expectations when --reboot shall reboot the system. The current code only reboots the system when the crashkernel parameter was updated during this run. But there are other opinions, that --reboot shall also reboot the system if a previous run of `kdumpctl reset-crashkernel` updated the crashkernel parameter and no reboot occurred yet. While possible this would add extra complexity to the code. Neither the confusion nor the extra complexity are justified, given that --reboot only replaces a single additional command for the user.

Thus deprecate the --reboot option so it can be removed later on.

### DIFF
--- a/crashkernel-howto.txt
+++ b/crashkernel-howto.txt
@@ -65,13 +65,13 @@ Reset using kdumpctl
 To make it easier to reset the `crashkernel=` kernel cmdline to this default
 value properly, `kdumpctl` also provides a sub-command:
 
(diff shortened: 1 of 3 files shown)


licliu:
How about always rebooting the system when the `--reboot` option is given? I think the user should have the expectation that the system will reboot when using this option.

coiby (approved):

licliu (approved):

Issue #35, opened 2024-09-18
Kdump service fails to start if the crashkernel command-line argument is absent in fadump case


sourabhjains:
Fadump reserves memory regardless of whether the crashkernel parameter is passed to the production kernel. Therefore, the expectation is that the kdump service should not depend on the crashkernel command-line argument when fadump is being used.

The patch that made the kdump service dependent on the crashkernel kernel command-line parameter is: https://github.com/rhkdump/kdump-utils/commit/0084806493d01d14dc7148b2dd5fc46e07b42c96.

The above commit did not take into account that fadump uses the kdump service, but crashkernel is not mandatory for fadump to reserve memory.

Any suggestions on how to remove the dependency on the crashkernel kernel parameter for the kdump service when fadump is active?

sourabhjains:
Any input on how to fix this issue?

Didn't find a way to restrict ConditionKernelCommandLine condition for only kdump.

liutgnu:
Hi @sourabhjains ,

Thanks for the issue, currently the maintainers are all in holiday until Oct 7. So it may take a few days before their reply. By a quick glimpse, I think it is a bug and should be fixed. I will take it a look tomorrow and try to give it a shot.

Thanks,
Tao Liu

liutgnu:
I guess what we want, is something like:

ConditionKernelCommandLine=crashkernel|fadump

But it is hard to implement the "OR" operation for the 2 strings. Though chatgpt can give a complicated way to implememt OR conditions.

I don't have further suggestions, see how other people think of it, or we just revert the 0084806 commit?

sourabhjains:
Hello @liutgnu 

Thanks for providing input on the issue.

> I guess what we want, is something like:
> 
> ConditionKernelCommandLine=crashkernel|fadump
> 
> But it is hard to implement the "OR" operation for the 2 strings. Though chatgpt can give a complicated way to implememt OR conditions.

Yes there is no native support to add conditions to ConditionKernelCommandLine. 

> 
> I don't have further suggestions, see how other people think of it, or we just revert the [0084806](https://github.com/rhkdump/kdump-utils/pull/39) commit?

Reverting https://github.com/rhkdump/kdump-utils/commit/0084806493d01d14dc7148b2dd5fc46e07b42c96 is fine because kdump scripts has checks for memory reservation.

But we care early kdump service exit if crashkernel is not present in kernel command-line then how about 
https://github.com/rhkdump/kdump-utils/pull/39  ?

PR #39, opened 2024-10-04
kdump.service: Replace ConditionKernelCommandLine with ExecCondition

sourabhjains:
Commit 0084806493d0 ("kdump.service: use ConditionKernelCommandLine=crashkernel") added a condition based on the crashkernel kernel command line parameter to control the start of the kdump service.

While ConditionKernelCommandLine=crashkernel works well for kdump, it causes issues for fadump (specific to the PowerPC architecture), which also uses the same service unit. Unlike kdump, crashkernel kernel command-line is not mandatory for fadump.

Since ConditionKernelCommandLine doesn't support evaluating multiple kernel command line parameters with dependencies between them, it has been replaced with ExecCondition to resolve this limitation.

Now, if fadump is configured and the crashkernel parameter is NOT present in the kernel command line, kdump service will start.

Fixes: #35

### DIFF
--- a/kdump.service
+++ b/kdump.service
@@ -2,10 +2,10 @@
 Description=Crash recovery kernel arming
 After=network.target network-online.target remote-fs.target basic.target
 DefaultDependencies=no
-ConditionKernelCommandLine=crashkernel
 
 [Service]
 Type=oneshot
+ExecCondition=/bin/sh -c 'grep -q -e "crashkernel" -e "fadump" /proc/cmdline'
 ExecStart=/usr/bin/kdumpctl start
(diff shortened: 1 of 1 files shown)


pfliu:
LGTM

coiby:
Hi @pfliu, can you also approve this PR using GitHub's UI interface? Thanks!

licliu (approved):
LGTM

pfliu (approved):

sourabhjains:
@pfliu  @licliu  Thanks for the review. @coiby , appreciate accepting the patch.

PR #41, opened 2024-10-08
kdump-lib-initramfs: Improve mount point retrieval logic

licliu:
We use findmnt to find the real mount point by mount source, however, when there is a bind mount target, findmnt will give more than one results and what we actually want is the one without fsroot.

Will fallback to previous method if the $_mntpoint is empty.

### DIFF
--- a/kdump-lib-initramfs.sh
+++ b/kdump-lib-initramfs.sh
@@ -103,8 +103,14 @@ get_fs_type_from_target()
 
 get_mntpoint_from_target()
 {
-	# --source is applied to ensure non-bind mount is returned
-	get_mount_info TARGET source "$1" -f
+	local _mntpoint
+	# get the first TARGET when SOURCE doesn't end with ].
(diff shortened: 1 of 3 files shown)


licliu:
The #37 btrfs part is not ready yet, but I  think this patch is ready for review.

coiby:
Hi @licliu,

1. You need to change get_bind_mount_source so it will call the new `get_mntpoint_from_target` as is the same to #28.
2. Can you add unit tests for get_mntpoint_from_target (you can tweak https://github.com/licliu/kdump-utils/pull/1/commits/d548389a535daaaf8195c46cedd1919b5278a06e for this purpose) to prevent possible regression? You may also want to add some code comments for this function as well.

licliu:
Thanks @coiby ! I'll post my v2 patch after adding unit tests.

licliu:
Hi @coiby , the v2 patch is posted, please kindly help to review.

coiby (approved):
LGTM, thanks!

Issue #46, opened 2024-10-14
Regression - ssh : bug parsing the ssh key option


jbtrystram:
kdump-utils-1.0.47-1 fails on remote ssh export. 

Using the following config : 

```
          ssh core@10.0.2.2
          sshkey /root/.ssh/id_ssh_kdump
          path /home/core/crash
          core_collector makedumpfile -F -l --message-level 1 -d 31
```

When triggering a kernel panic : 
```
         Starting kdump-capture.service - Kdump Vmcore Save Service...
[    2.308392] kdump[642]: Kdump is using the default log level(3).
[    2.239662] kdump[642]: Kdump is using the default log level(3).
[    3.344918] kdump[694]: saving to core@10.0.2.2:/home/core/crash/10.0.2.15-2024-10-14-13:01:14
[    3.276325] kdump[694]: saving to core@10.0.2.2:/home/core/crash/10.0.2.15-2024-10-14-13:01:14
[    3.351601] kdump.sh[696]: Warning: Identity file  /root/.ssh/id_ssh_kdump -o BatchMode=yes -o StrictHostKeyChecking=yes not accessible: No such file or directory
.
[    3.284293] kdump.sh[696]: Warning: Identity file  /root/.ssh/id_ssh_kdump -o BatchMode=yes -o StrictHostKeyChecking=yes not accessible: No such file or directory
.
[    3.375233] kdump[701]: saving vmcore status file to /vmcorestatus//var/crash/vmcore-creation.status
[    3.306658] kdump[701]: saving vmcore status file to /vmcorestatus//var/crash/vmcore-creation.status
[    3.378655] kdump[705]: saving vmcore failed
[    3.309817] kdump[705]: saving vmcore failed
[    3.380191] systemd[1]: kdump-capture.service: Main process exited, code=exited, status=1/FAILURE
[    3.311558] systemd[1]: kdump-capture.service: Main process exited, code=exited, status=1/FAILURE
[    3.381481] systemd[1]: kdump-capture.service: Failed with result 'exit-code'.
[    3.312776] systemd[1]: kdump-capture.service: Failed with result 'exit-code'.
[FAILED] Failed to start kdump-capture.service - Kdump Vmcore Save Service.
See 'systemctl status kdump-capture.service' for details.

```

Looking at this line : `Warning: Identity file  /root/.ssh/id_ssh_kdump -o BatchMode=yes -o StrictHostKeyChecking=yes not accessible: No such file or directory`
[...]

PR #47, opened 2024-10-15
Allow ssh opts to be processed correctly

coiby:
Resolves: https://github.com/rhkdump/kdump-utils/issues/46

ssh opts shouldn't be quoted otherwise it will be treated as one argument. For clarity, also rename ssh_opt to ssh_opts.

Fixes: 7218f584 ("99kdumpbase: fix shellcheck warnings")
Reported-by: Jean-Baptiste Trystram <jbtrystram@redhat.com>

### DIFF
--- a/dracut/99kdumpbase/kdump.sh
+++ b/dracut/99kdumpbase/kdump.sh
@@ -396,7 +396,7 @@ dump_raw() {
 # $2: ssh address in <user>@<host> format
 dump_ssh() {
     _ret=0
-    _ssh_opt="-i $1 -o BatchMode=yes -o StrictHostKeyChecking=yes"
+    _ssh_opts="-i $1 -o BatchMode=yes -o StrictHostKeyChecking=yes"
     _ssh_dir="$KDUMP_PATH/$HOST_IP-$DATEDIR"
     if is_ipv6_address "$2"; then
         _scp_address=${2%@*}@"[${2#*@}]"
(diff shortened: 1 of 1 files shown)


coiby:
Hi @prudo1 and @licliu,

Since [bz2284097](https://bugzilla.redhat.com/show_bug.cgi?id=2284097) is an AcceptedFreezeException bug and affected by this issue, so I just quickly propose a solution first.

licliu:
Hi @coiby, I think the `$2 ` in this [line](https://github.com/rhkdump/kdump-utils/blob/6f32fab791746da3f075273e156bd70055db5d4c/dracut/99kdumpbase/kdump.sh#L464) is `_ssh_opts` and this patch doesn't remove the quotes.

coiby:
> Hi @coiby, I think the `$2 ` in this [line](https://github.com/rhkdump/kdump-utils/blob/6f32fab791746da3f075273e156bd70055db5d4c/dracut/99kdumpbase/kdump.sh#L464) is `_ssh_opts` and this patch doesn't remove the quotes.

Good catch! This issue has been fixes in new version. Thanks!

dracut/99kdumpbase/kdump.sh
```diff
@@ -461,13 +467,16 @@ save_opalcore_ssh() {
 
     dinfo "saving opalcore:$OPALCORE to $3:$1"
 
+    # shellcheck disable=SC2086 # $2 (_ssh_opts) needs to be split
     if ! scp "$2" "$OPALCORE" "$4:$1/opalcore-incomplete"; then
```
prudo1, line 471: Hi @coiby,
thanks for taking care of this. I believe you missed the quotes on the line above. Other than that the commit looks fine. Even when having all the checks disabled isn't really ideal.
BTW, when I looked at the problem yesterday I got a little bit side tracked and rewrote `dump_ssh` to use `curl` rather than `ssh/scp`. So far it seems to work all fine and it  cleaned up the code quite a bit. Do you think it makes sense to go that direction?

prudo1 (changes requested):

dracut/99kdumpbase/kdump.sh
coiby: Hi @prudo1,

Oh yes, I missed the quotes. It's fixed in new version, thanks! 

If there is a more elegant solution, that couldn't be better. Can you send a draft patch? I'm also curious to know curl could provide a different solution.

prudo1 (approved):
Looks fine. Thanks!

My code to use curl is still a little messy but I'll clean it up and send a RFC.

PR #48, opened 2024-10-18
Return the correct exit code of rebuild initrd

liutgnu:
Resolves: https://issues.redhat.com/browse/RHEL-63047
    
The exit code of rebuild_initrd() should be either of
rebuild_kdump/fadump_initrd(), rather than set_vmcore_creation_status(),
otherwise it will cause a regression when rebuild initrd fails.
    
Fixes: 88525ebf ("Introduce vmcore creation notification to kdump")
    
### DIFF
--- a/kdumpctl
+++ b/kdumpctl
@@ -173,6 +173,7 @@ rebuild_kdump_initrd()
 
 rebuild_initrd()
 {
+	local _ret
 	dinfo "Rebuilding $TARGET_INITRD"
 
 	if [[ ! -w $(dirname "$TARGET_INITRD") ]]; then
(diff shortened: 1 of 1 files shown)


coiby:
Hi @liutgnu, the patch looks good to me except for a minor concern. Can you also add a reference to a bug i.e. a `Resolves:` tag?

liutgnu:
> Hi @liutgnu, the patch looks good to me except for a minor concern. Can you also add a reference to a bug i.e. a `Resolves:` tag?

Hi @coiby , Thanks for your comments and quick response. I just force pushed v2 for this.

PR #49, opened 2024-10-18
fadump: pass additional parameters for capture kernel

hbathini:
Since kernel commit 3416c9daa6b13 ("powerpc/fadump: pass additional parameters when fadump is active"), fadump supports passing additional parameters to dump capture kernel. Leverage that support here to pass additional parameters to dump capture kernel.

Also, update fadump-howto.txt to make clear on the options that are not relevant for fadump in /etc/sysconfig/kdump

The default bootargs to append for fadump capture kernel boot are chosen with the intent to optimize resources and reduce memory footprint in dump capture environment.

### DIFF
--- a/gen-kdump-sysconfig.sh
+++ b/gen-kdump-sysconfig.sh
@@ -31,6 +31,10 @@ KDUMP_COMMANDLINE_REMOVE="hugepages hugepagesz slub_debug quiet log_buf_len swio
 # after processed by KDUMP_COMMANDLINE_REMOVE
 KDUMP_COMMANDLINE_APPEND="irqpoll maxcpus=1 reset_devices novmcoredd cma=0 hugetlb_cma=0"
 
+# This variable lets us append arguments to fadump (powerpc) capture kernel,
+# further to the parameters passed via the bootloader.
+FADUMP_COMMANDLINE_APPEND=""
+
 # Any additional kexec arguments required.  In most situations, this should
 # be left empty
 #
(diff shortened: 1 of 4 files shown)


gen-kdump-sysconfig.sh
```diff
@@ -90,13 +94,17 @@ ppc64)
 		"hugepages hugepagesz slub_debug quiet log_buf_len swiotlb hugetlb_cma ignition.firstboot"
 	update_param KDUMP_COMMANDLINE_APPEND \
 		"irqpoll maxcpus=1 noirqdistrib reset_devices cgroup_disable=memory numa=off udev.children-max=2 ehea.use_mcs=0 panic=10 kvm_cma_resv_ratio=0 transparent_hugepage=never novmcoredd hugetlb_cma=0"
+	update_param FADUMP_COMMANDLINE_APPEND \
+		"nr_cpus=16 numa=off cgroup_disable=memory cma=0 kvm_cma_resv_ratio=0 hugetlb_cma=0 transparent_hugepage=never novmcoredd udev.children-max=2"
```
pfliu, line 98: Is ppc big-endian supported yet? Since maxcpus parameter is not change for that arch. So I prefer to keep ppc case unchanged.

Otherwise, LGTM.


Thanks,

Pingfan

pfliu (changes requested):

gen-kdump-sysconfig.sh
hbathini: Hello Pingfan, thanks for the review.
The pass additional parameters support applies to both big endian and little endian variants.
Also, nr_cpus support has always been there for fadump but there was no practical way to pass additional parameters to dump capture kernel for fadump till kernel commit 3416c9daa6b13 ("powerpc/fadump: pass additional parameters when fadump is active") made it possible..

pfliu: Hello Hari, my point is there is no support for ppc64be on either Fedora or RHEL. So the branch of ppc64be will be cleaned up later, and no new changes to it any more.

Thanks

hbathini: Ah! You mean, no changes to ppc64. Fine by me. Updated the pull request dropping the ppc64 hunk..

pfliu (approved):
It looks good to me now. Thanks Hari!

PR #53, opened 2024-10-25
Re-introduce vmcore creation notification to kdump

liutgnu:
Motivation
    ==========
    
    unexpected for kdump.
    
    It is highly recommended people to test kdump after any system modification,
    such as:
    
       firmware upgrading etc.
       etc.
    
    Design
    ======
    
    indicator of such modification, and kdump need to be tested. This will
    clear the vmcore creation status specified in $VMCORE_CREATION_STATUS,
    and as a result, a notification of vmcore creation test will be
    outputted.
    
    To test kdump, there is an entry for doing that by "kdumpctl test". It
    will generate a timestamp string as the ID of the current test, along
    with a "pending" status in $VMCORE_CREATION_STATUS, then a real crash &
    dump process will be triggered.
    
    After system reboot back to normal, a vmcore creation check will start at
    "kdumpctl (re)start/status", and will report the results as
    success/fail/manual status to users.
    
    To achieve that, program will first check the status in $VMCORE_CREATION_STATUS.
    If "pending" status if found, which means the test result is
    undetermined and need a retrive from remote/local dump folder. Then if test
    id is found in the dump folder and vmcore is complete, then "pending"
    would be overwritten by "success", which indicates a successful kdump
    test. If test id is found in the dump folder but vmcore is incomplete,
    then it is a "fail" kdump test. If no test id is found, then it is a "manual"
    status, which indicates users should check the test results manually.
    
    If $VMCORE_CREATION_STATUS is already success/fail/manual status, it indicates
    the test result has already been determined, so the program will not access
    the remote/local dump folder again. This can limite any unnecessary
    access to dump target, shorten the time consumption.
    
    User should check for the root cause of fail/manual status when get
    reports.
    
    the current env. The format is like:
    
       <status> kdump_test_id=<timestamp sec>-<timestamp nanosec>
    e.g:
       success kdump_test_id=1729823462-938751820
    
    Which means, there has been a successful kdump test at
    $(date -d "@1729823462") timestamp for the current env. Timestamp
    nanosec is only meaningful for uniquify id string.
    
    Difference
    ==========
    Previously there is one commit 88525ebf ("Introduce vmcore creation
    notification to kdump") merged and addressing the same issue, but
    implemented differently:
    
    The prev one:
    Save the $VMCORE_CREATION_STATUS to local drive during the 2nd kernel
    dumping. If vmcore dumping target is different from $VMCORE_CREATION_STATUS's
    drive, then the latter one need to be mounted in 2nd kernel.
    
    This one:
    Save the $VMCORE_CREATION_STATUS to local drive only in 1nd kernel, that
    is, the test result is retrived after 2nd kernel dumping. So it doesn't
    load or mount other drive in 2nd kernel.
    
    The advantage:
    Extra mounting in 2nd kernel will introduce higher risk of failure,
    as a result, lower the success of vmcore dumping, which is
    unaccepted. So keep the code for 2nd kernel as simple is preferred.
    
    Usage
    =====
    kdump: Stopping kdump: [OK]
    kdump: Starting kdump: [OK]
    kdump: Notice: No vmcore creation test performed!
    
    kdump: Kdump is operational
    kdump: Notice: No vmcore creation test performed!
    
    [root@localhost ~]# kdumpctl test
    
    [root@localhost ~]# cat /var/lib/kdump/vmcore-creation.status
    pending kdump_test_id=1729823462-938751820
    
    kdump: Kdump is operational
    kdump: Notice: Last successful vmcore creation on Fri Oct 25 02:31:02 AM UTC 2024
    
    [root@localhost ~]# cat /var/lib/kdump/vmcore-creation.status
    success kdump_test_id=1729823462-938751820
    
    kdump: Stopping kdump: [OK]
    kdump: Starting kdump: [OK]
    kdump: Notice: Last successful vmcore creation on Fri Oct 25 02:31:02 AM UTC 2024
    
    Note: the notification for kdumpctl (re)start/status can be disabled by
    setting VMCORE_CREATION_NOTIFICATION in /etc/sysconfig/kdump. And fadump
    is NOT supported for this feature.
    
### DIFF
--- a/kdumpctl
+++ b/kdumpctl
@@ -14,6 +14,7 @@ KDUMP_INITRD=""
 TARGET_INITRD=""
 #kdump shall be the default dump mode
 DEFAULT_DUMP_MODE="kdump"
+VMCORE_CREATION_STATUS="/var/lib/kdump/vmcore-creation.status"
 
 # Some default values in case /etc/sysconfig/kdump doesn't include
 KDUMP_COMMANDLINE_REMOVE="hugepages hugepagesz slub_debug"
@@ -45,8 +46,10 @@ if ! dlog_init; then
 fi
 
 KDUMP_TMPDIR=$(mktemp --tmpdir -d kdump.XXXX)
+TMPMNT="$KDUMP_TMPDIR/target"
 trap '
     ret=$?;
+    is_mounted $TMPMNT && umount -f $TMPMNT;
     rm -rf "$KDUMP_TMPDIR"
     exit $ret;
     ' EXIT
@@ -185,6 +188,11 @@ rebuild_initrd()
 	else
 		rebuild_kdump_initrd
 	fi
+
+	_ret=$?
+
+	set_vmcore_creation_status 'clear'
+	return $_ret
 }
 
 #$1: the files to be checked with IFS=' '
@@ -1674,6 +1682,170 @@ _should_reset_crashkernel() {
 	[[ $(kdump_get_conf_val auto_reset_crashkernel) != no ]] && systemctl is-enabled kdump &> /dev/null
 }
 
+set_kdump_test_id()
+{
+	local _id=$1
+
+	KDUMP_COMMANDLINE_APPEND+=" $_id "
+	reload >& /dev/null
+
+	if [[ "$?" -ne 0 ]]; then
+		derror "Set kdump test id fail."
+		exit 1
+	fi
+}
+
+# $1: success/fail/pending/manual/clear
+# $2: test id
+set_vmcore_creation_status()
+{
+	local _status=$1
+	local _kdump_test_id
+	_dir=$(dirname "$VMCORE_CREATION_STATUS")
+
+	[[ -d "$_dir" ]] || mkdir -p "$_dir"
+	[[ -w "$_dir" ]] || chmod +w "$_dir"
+
+	case "$_status" in
+		pending)
+			_kdump_test_id="kdump_test_id=$(date +%s-%N)"
+			set_kdump_test_id "$_kdump_test_id"
+			echo  "$_status $_kdump_test_id" > "$VMCORE_CREATION_STATUS"
+			;;
+		success | fail | manual)
+			sed -E -i "s/^\w+/$_status/" "$VMCORE_CREATION_STATUS"
+			;;
+		clear)
+			rm -f "$VMCORE_CREATION_STATUS"
+			;;
+		*)
+			return
+	esac
+	sync -f "$_dir"
+}
+
+fetch_status()
+{
+	local _test_id="$1" _mnt
+	local  _status
+
+	is_raw_dump_target && return 2
+
+	_status="${OPT[path]}/kdump-test-$_test_id/vmcore-creation.status"
+
+	if is_nfs_dump_target || is_local_target; then
+		_mnt=$(get_mntpoint_from_target "${OPT[_target]}")
+		if [[ -z "$_mnt" ]] || ! is_mounted "$_mnt"; then
+			mkdir -p $TMPMNT
+			mount "${OPT[_target]}" "$TMPMNT" -t "${OPT[_fstype]}" -o defaults || \
+				{ dwarn "Failed to mount ${OPT[_target]}" && return 2; }
+			_mnt="$TMPMNT"
+		fi
+		_status="$_mnt/$_status"
+	elif is_ssh_dump_target; then
+		scp -i "${OPT[sshkey]}" -o BatchMode=yes \
+			"${OPT[_target]}:$_status" \
+			"$KDUMP_TMPDIR"
+		case "$?" in
+			0)
+				# success
+				;;
+			1)
+				# file not found
+				return 1
+				;;
+			255)
+				# no connection to host
+				return 2
+		esac
+		_status="$KDUMP_TMPDIR/vmcore-creation.status"
+	fi
+
+	[[ -f "$_status" ]] || return 1
+	grep -q "success" "$_status" && return 0 || return 1
+}
+
+check_vmcore_creation_status()
+{
+	local _status _test_id _timestamp _status_date
+
+	[[ ${VMCORE_CREATION_NOTIFICATION,,} == "yes" ]] || return
+
+	[[ "$DEFAULT_DUMP_MODE" == "kdump" ]] || return
+
+	if [[ ! -s "$VMCORE_CREATION_STATUS" ]]; then
+		dwarn "Notice: No vmcore creation test performed!"
+		return
+	fi
+
+	[[ "${#OPT[@]}" -eq 0 ]] && { parse_config || return; }
+
+	read -r _status _test_id < "$VMCORE_CREATION_STATUS"
+	_test_id=${_test_id#*=}
+	_timestamp=${_test_id%-*}
+	_status_date=$(date -d "@$_timestamp")
+
+	if [[ "$_status" == "pending" ]]; then
+		fetch_status "$_test_id"
+		case "$?" in
+		0)
+			_status="success"
+			;;
+		1)
+			_status="fail"
+			;;
+		*)
+			_status="manual"
+			;;
+		esac
+		set_vmcore_creation_status "$_status"
+	fi
+
+	case "$_status" in
+		success)
+			dinfo "Notice: Last successful vmcore creation on $_status_date"
+			;;
+		fail)
+			dwarn "Notice: Last NOT successful vmcore creation on $_status_date"
+			;;
+		manual)
+			dwarn "Notice: Require manual check for kdump test of $_status_date"
+			;;
+		*)
+			derror "Unknown test status: $_status"
+			;;
+	esac
+}
+
+kdump_test()
+{
+	if ! is_kernel_loaded "$DEFAULT_DUMP_MODE"; then
+		derror "Kdump needs be operational before test."
+		exit 1
+	fi
+
+	if [[ ! "$DEFAULT_DUMP_MODE" == "kdump" ]]; then
+		derror "Only kdump is supported for test."
+		exit 1
+	fi
+
+	if [[ ! "$1" == "--force" ]]; then
+		read -p "DANGER!!! Will perform a kdump test by crashing the system, proceed? (y/N): " input
+		case $input in
+		[Yy] )
+			dinfo "Start kdump test..."
+			;;
+		* )
+			dinfo "Operation cancelled."
+			exit 0
+			;;
+		esac
+	fi
+
+	set_vmcore_creation_status 'pending'
+	echo c > /proc/sysrq-trigger
+}
+
 main()
 {
 	# Determine if the dump mode is kdump or fadump
@@ -1705,6 +1877,7 @@ main()
 			EXIT_CODE=3
 			;;
 		esac
+		check_vmcore_creation_status
 		exit $EXIT_CODE
 		;;
 	reload)
--- a/kdumpctl
+++ b/kdumpctl
@@ -173,7 +173,6 @@ rebuild_kdump_initrd()
 
 rebuild_initrd()
 {
-	local _ret
 	dinfo "Rebuilding $TARGET_INITRD"
 
 	if [[ ! -w $(dirname "$TARGET_INITRD") ]]; then
@@ -186,10 +185,6 @@ rebuild_initrd()
 	else
 		rebuild_kdump_initrd
 	fi
-	_ret=$?
-
-	set_vmcore_creation_status 'clear' "$VMCORE_CREATION_STATUS"
-	return $_ret
 }
 
 #$1: the files to be checked with IFS=' '
@@ -1064,7 +1059,6 @@ start()
 	start_dump || return
 
 	dinfo "Starting kdump: [OK]"
-	check_vmcore_creation_status
 	return 0
 }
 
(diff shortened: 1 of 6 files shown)


kdumpctl
```diff
@@ -1659,47 +1662,156 @@ _should_reset_crashkernel() {
...
-		derror "Vmcore status dir $_dir not exist."
-		exit 1
-	fi
-
-	if ! lsblk $(get_mount_info SOURCE target "$_dir") > /dev/null; then
-		derror "$VMCORE_CREATION_STATUS must on local drive"
-		exit 1
-	fi
-
 	if [[ ! "$1" == "--force" ]]; then
```
licliu, line 1815: @liutgnu There is a $1, but seems no arguments are passed.

kdumpctl
```diff
@@ -1794,7 +1907,7 @@ main()
 		;;
 	test)
 		shift
-		kdump_test "$@"
+		kdump_test
```
licliu, line 1910: maybe missed $@ here?

liutgnu: Yes, thanks for pointing it out, it should be "$@". I will fix it in v2.

kdumpctl
liutgnu: Correct, the argument is missing, "$@" need to passed.

kdumpctl
```diff
@@ -1653,6 +1662,173 @@ _should_reset_crashkernel() {
...
+	esac
+	sync -f "$_dir"
+}
+
+fetch_test_status()
+{
+	local _test_id="$1"
+	local _script="$KDUMP_TMPDIR/fetch.sh"
+
+	cat << EOF > "$_script"
```
coiby, line 1715: Hi @liutgnu, why do you create a temporary script `$KDUMP_TMPDIR/fetch.sh` first? Can't it run as a function?

liutgnu: The script is for iterate the target folder for kexec-dmesg.log and vmcore-dmesg.txt which holding the test_id string. For local iteration, the script can be simplified into a function, but for ssh dump, I didn't find a way to execute a function via ssh, but only create a script file and ask ssh to execute it remotely.Do you have better solutions for that?

coiby: A function won't work for ssh dump, thanks for enlightening me! It seems there is no need for variable expansion in fetch. You can simply use `cat << \EOF > "$_script"` to avoid escaping `$1` thus to increase code readability. If `fetch.sh` get larger, it's better not to generate on the fly so shellcheck can do some static analysis. 

Btw, why `--` is used in `ssh -i "${OPT[sshkey]}" -o BatchMode=yes "${OPT[_target]}" 'bash -s' -- < "$_script" "${OPT[path]}"`?

kdumpctl
```diff
@@ -1659,47 +1662,156 @@ _should_reset_crashkernel() {
...
+		;;
+esac
+if [[ -f "\$dir/vmcore" || -f "\$dir/vmcore.flat" ]]; then
+	# Success
+	exit 0
+else
+	# Fail
+	exit 1
+fi
+EOF
```
prudo1, line 1748: I'm not really comfortable with creating and running the script as is. For one it creates a dependency for the host, when using a ssh target. For the other writing the script to a temporary file and then running it opens a race window in which attackers could replace the script with their own. Leading to a potential privilege escalation when we don't pay enough attention.

Could we simplify this by storing the status in a file known to both kdumpctl and the initrd, e.g. something like `$path/.$test_id.status`. That would require to run some code in the initrd to parse the `$test_id` from the command line and write the final status to the file. But that would allow to make the code overall much simpler IMHO and still prevents the need to always mount the rootfs in the initrd.

kdumpctl
```diff
@@ -1659,47 +1662,156 @@ _should_reset_crashkernel() {
...
+	if is_nfs_dump_target || is_local_target; then
+		mkdir -p "$TMPMNT"
+		mount "${OPT[_target]}" "$TMPMNT" -t "${OPT[_fstype]}" -o defaults || \
+			{ dwarn "Failed to mount ${OPT[_target]}" && return 2; }
+		bash "$_script" "$TMPMNT/${OPT[path]}"
+	elif is_ssh_dump_target; then
```
prudo1, line 1754: I believe we should only mount the target when it isn't mounted already.

kdumpctl.8
```diff
@@ -70,16 +70,6 @@ Note: The memory requirements for kdump varies heavily depending on the
...
-.TP
-.I test [--force]
-Test the kdump by actually trigger the system crash & dump, and check if a
-vmcore can really be generated successfully based on current config and
-environment. After system reboot back to normal, check the test result
-by "kdumpctl status".
-
-If the optional parameter [--force] is provided, there will be no interact
-before triggering the system crash. Dangerous though, this option is meant
-for automation testing.
```
prudo1, line 73: This hunk shouldn't be reverted.

prudo1 (changes requested):
Hi @liutgnu,
overall this version looks much better. But I'm a little bit concerned about using the generated script to fetch the status. Anyway, please see my comments below.

kdumpctl
liutgnu: I will try to get rid of the temporarty script file, as Philipp pointed out, the script file can be hacked on the disk, which may leave security issues. No script may be better. "bash -s -- < script argument" will ask bash to execute the script with the argument passed to the script, seems the same as "bash -s < script argument"...

kdumpctl
```diff
@@ -1659,47 +1662,156 @@ _should_reset_crashkernel() {
...
+	local _test_id=$2
+	_dir=$(dirname "$VMCORE_CREATION_STATUS")
+
+	[[ -d "$_dir" ]] || mkdir -p "$_dir"
+	[[ -w "$_dir" ]] || chmod +w "$_dir"
+
+	case "$_status" in
+		pending)
+			echo  "$_status $_test_id" > "$VMCORE_CREATION_STATUS"
+			;;
```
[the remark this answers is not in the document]
liutgnu: Yes, agreed, will update it.

kdumpctl
liutgnu: The reason I didn't check the mount is, it seems not a big issue to me if the drive is already mounted before, and we mount the same drive to a different place again. Since a) we are in 1st kernel here, memory usage is not a problem, and b) we only read into the drive, no write operations, so it is harmless, and c) extra code will be involved for mount checking and switching to the mounting, seems redundant and unecessary. What do you think?

kdumpctl.8
liutgnu: Agreed, will fix it.

kdumpctl
```diff
@@ -14,6 +14,7 @@ KDUMP_INITRD=""
 TARGET_INITRD=""
 #kdump shall be the default dump mode
 DEFAULT_DUMP_MODE="kdump"
+VMCORE_CREATION_STATUS="/var/crash/vmcore-creation.status"
```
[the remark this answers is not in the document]
daveyoung: > Hi Philipp,
> 
> It doesn't matter what path the user set in kdump.conf, we just save the status file into /var/crash and done, as long as we can r/w it. I have no objection for "/var/lib/kdump", but I'm a little worry if selinux will prevent such writing, since for /var/crash, we already have the write permission, but as for "/var/lib/kdump", maybe we need to update the selinux rules as well. I haven't tested it though...

Hi @liutgnu ,  maybe no need to worry about  as selinux has fixed a issue here:
https://bugzilla.redhat.com/show_bug.cgi?id=1965985
https://github.com/fedora-selinux/selinux-policy/commit/af7e4b6492b315ee0912aef7f9f2d89a4f681ab6

> 
> Thanks, Tao Liu

kdumpctl
liutgnu: Hi Philipp,

> I'm not really comfortable with creating and running the script as is. For one it creates a dependency for the host, when using a ssh target. For the other writing the script to a temporary file and then running it opens a race 

I'm not very clear for the dependency here, do you suggest we don't ask remote server to execute script at all?

> window in which attackers could replace the script with their own. Leading to a potential privilege escalation when we don't pay enough attention.

Agreed, we can get rid of the temporary script file, how about a use of typeset like this: https://stackoverflow.com/questions/22107610/shell-script-run-function-from-script-over-ssh?

> 
> Could we simplify this by storing the status in a file known to both kdumpctl and the initrd, e.g. something like `$path/.$test_id.status`. That would require to run some code in the initrd to parse the `$test_id` from the command line and write the final status to the file. But that would allow to make the code overall much simpler IMHO and still prevents the need to always mount the rootfs in the initrd.

If we want initrd, aka 2nd kernel to access the status file, assuming the status file is independent from vmcore it  is trying to save, this will involve extra kernel module to deal with the fs where the status file is located, and this is what we are trying to avoid, because extra kernel module will increase the kernel failure, preventing vmcore saving, and it is unacceptable.

If the status file is saved adjoining vmcore, then no extra module is needed. Since we still need to access the remote server to check for the status when we back into 1st kernel, so there is no point for creating the status file next to vmcore remotely right?

Please correct me if I misunderstood your approach...

kdumpctl
liutgnu: See v2 which fixed the bug.

kdumpctl
liutgnu: See v2, which I removed the temporary script file to fetch test results.

kdumpctl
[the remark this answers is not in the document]
liutgnu: Hi @daveyoung & @prudo1 ,

Thanks for the info, please see v2 which I changed "/var/crash/vmcore-creation.status" to "/var/lib/kdump/vmcore-creation.status".

kdumpctl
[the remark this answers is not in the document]
liutgnu: I have moved the kdump_test_id call into the pending) here by v2.

kdumpctl.8
liutgnu: restored the doc file in v2.

kdumpctl
liutgnu: In v2, I checked if the target is mounted or not before mount it into tmpmnt. So the device will only mount once.

kdumpctl
liutgnu: I removed the temporary script generation in v2.

kdumpctl
```diff
@@ -1672,47 +1675,162 @@ _should_reset_crashkernel() {
...
 }
 
+set_kdump_test_id()
+{
+	local _id=$1
+	local _kmsg="/dev/kmsg"
+
+	if [[ $DEFAULT_DUMP_MODE == "fadump" ]]; then
+		[[ -c "$_kmsg" ]] && echo "$_id" > "$_kmsg"
+	else
```
prudo1, line 1685: I noticed that @hbathini just added a mechanism to append parameters to the fadump kernel command line in
77b80ce ("fadump: pass additional parameters for capture kernel"). Using that would make this feature much simpler as well. As then kdump and fadump can be handled the same way.

Especially it would make it trivial to parse the `kdump_test_id` from the command line within the initrd for both kdump and fadump. That's why I believe it would make the whole implementation simpler and less error prone if some of the functionality is moved to the initrd.

What I have in mind is:
1) `kdumpctl test` sets the `kdump_test_id` like in your patch.
2) the initrd parses the `kdump_test_id` from the command line and creates a temporary `$path/.$kdump_test_id.status' file to write the status to. As the initrd knows when dumping has succeeded it could simply initialize the file to 'fail' and update it to 'success' right before calling the final_action.
3) When `kdumpctl` fetches the status it only needs to check whether this file exists and then can simply read the status from it without parsing a whole bunch of `dmesg` files.

The only "problem" I see is that `kdump.sh` probably will need a small refractoring/cleanup for ssh dumps. But that is probably something we should do anyway.

What do you think?

liutgnu: Hi @prudo1 ,

Thanks for the comments! I have no objection for your approach itself, however there is only one concerns from my side, the fadump support of additional parameter for 2nd kernel is a new feature, which requires kernel patch (3416c9daa6b13 "powerpc/fadump: pass additional parameters when fadump is active") etc backporting. I'm not sure whether the kernel patch will be backported to rhel9 or rhel8 or some older version of rhel9.1/2/3..., since CEE requires the "kdumpctl test" feature mainly to support older kernels in their daily work, and I guess if CEE think the "kdumpctl test" feature is good enough, they probably would ask us to support it in rhel8 as well.

If they do, this will make a burden for us to maintain both traditional & newer fadump. Dave suggested me to ask if CEE wants "kdumpctl test" for fadump, because comparing to kdump, fadump is more stable and probably don't need "kdumpctl test" at all. Before CEE comfirmation, I would suggest to keep "kdumpctl test" minimal dependency for easier backporting, what do you think?

liutgnu: Other than passing the kdump_test_id string through command line to 2nd kernel, I cannot think of a better way to do that. Packing the kdump_test_id string into initrd img as a file is also not a good idea, this will ask initrd rebuild every time for kdumpctl test. The rebuilt initrd will not reflect the true env of what we are trying to test.

Or 2nd kernel doesn't have to know kdump_test_id at all, just save the vmcore as it used to do. After vmcore saved success, it can set a mark as success/failure next to vmcore. When back to 1st kernel, we can fetch for the mark. But this approach have no difference with current approach, vmcore itself is the success/failure mark already...

prudo1: Hi @liutgnu,

you are right, that depending on the kernel patch isn't ideal but I wouldn't worry about it too much. The way I see it powerpc and fadump are only used by a small number of our users. So the question is if it is worth adding so much complexity, which we need to support in the future, only because a small number of users might need it for older releases. Especially as the kernel commit in question has already been backported to c9s. So all future releases will support it. Personally I'm tending more to keep the implementation simple and not support fadump on older releases.
And in case CEE (or whoever) thinks that fadump support is needed even for older releases and get's the exception to include it. Then we still can find a way to get it to work there. I just don't think that we should include it into the 'standard' implementation we'll need to support for the next 10+ years.

True, there's no suitable alternative other than passing it via the command line. That's why I don't think it makes sense to parse all the dmesg files to find the id when there is all the infrastructure in place to parse the id from the command line in the initrd and use it to simplify the way we report the result back to the production system.

liutgnu: Hi @prudo1 ,

Thanks for the comments, agreed. In addition, CEE replied to me that they don't need "kdumpctl test" for fadump, so I removed fadump related part in v3, please check it out. Thanks!

liutgnu:
Hi @prudo1 ,

I cherry-picked the "kdumpctl test" patch from your repo, and made some change against it, because your repo had some clean up and refactoring on $KDUMP_PATH. The code change is dealing with the lack of refactoring. Please check out the v4. Thanks!

prudo1 (approved):
Hi @liutgnu,

all in all looks good to me although not having the cleanups makes the code in `kdump.sh` pretty ugly. But that's something we can cleanup later.

I gave the issue an other thought over the weekend and found some small nits/possible improvements. But they aren't critical and more food for thought for the future.
1. We only need to track the successful cases in the initrd as we treat having no status file as failure when fetching the result. So we could drop setting the status to failure in `kdump_test_init` and with that simplify `kdump_test_set_status`. I.e. we can assume that the target directory has already been created (if not `dump_{fs,ssh}` would have returned with an error already)
2. One downside of saving the status file in the same directory as the dump is that overeager admins might delete it together with the dump before the status was fetched. Thus leading to a false-negative result. From that POV it would make sense to move the status file up one directory. This however has the downside that manually verifying the status is more difficult. Not sure what the better solution is.
3. In general I think we should rethink on when we set the status to `manual`. Especially as the current code does not allow to re-fetch the status. For example when the dump device cannot be mounted. Wouldn't it make more sense to fail with an error message telling the user to mount the device manually and re-run `kdumpctl status`? Similar when the ssh host cannot be reached.

liutgnu:
Hi @prudo1,

Thanks for your help on the code review and improvement!

> Hi @liutgnu,
> 
> all in all looks good to me although not having the cleanups makes the code in `kdump.sh` pretty ugly. But that's something we can cleanup later.
> 
> I gave the issue an other thought over the weekend and found some small nits/possible improvements. But they aren't critical and more food for thought for the future.
> 
> 1. We only need to track the successful cases in the initrd as we treat having no status file as failure when fetching the result. So we could drop setting the status to failure in `kdump_test_init` and with that simplify `kdump_test_set_status`. I.e. we can assume that the target directory has already been created (if not `dump_{fs,ssh}` would have returned with an error already)

Please refer to the 3rd one for my understanding of manual state.

> 2. One downside of saving the status file in the same directory as the dump is that overeager admins might delete it together with the dump before the status was fetched. Thus leading to a false-negative result. From that POV it would make sense to move the status file up one directory. This however has the downside that manually verifying the status is more difficult. Not sure what the better solution is.

Yes indeed. I guess we can go as it is. If in the future people report that false-negative result is a problem we can get it fixed later. Frankly I haven't receive any feedback from CEE about "kdumpctl test" feature, we can get it done then.

> 3. In general I think we should rethink on when we set the status to `manual`. Especially as the current code does not allow to re-fetch the status. For example when the dump device cannot be mounted. Wouldn't it make more sense to fail with an error message telling the user to mount the device manually and re-run `kdumpctl status`? Similar when the ssh host cannot be reached.

My original thought about the success/fail/manual is, for success/fail, is the state that we get confirmed; for manual, is the state we cannot confirm by our program. In my previous implementation, the success is set when "test id string found && vmcore is found"; fail is when "test id string found && vmcore not found"; the other cases are manual, which are "test id string not found". You can see the "manual" status covers cases wider that the success/fail, including "mount remote storage fail so cannot fetch test result". Because solving network/mount errors need users intervention, that's indeed "manual" right? And the test result checked by user himself is much reliable than we do, because there are a lot of reasons for "kdumpctl test" fails, from a tiny mistake to some serious faulty, users should check for that. Our script is just "try our best".

In this case, if the status is set to "manual" but turns out to be a tiny mistake, such as "No ip address assigned", users can fix the simple network issue, reset the status file to be "pending" again, and retrigger the "kdumpctl status", our program can "re-fetch" the results.

What do you think?

PR #57, opened 2024-11-18
fadump: fix passing additional parameters for capture kernel

hbathini:
Currently, additional parameters for fadump capture kernel are being set only with reload command. In fact, the check for bootargs_append node before writing to it is incorrect too. Fix it and also, setup fadump aditional parameters for start/restart case as well.

Fixes: 77b80ce5 ("fadump: pass additional parameters for capture kernel")

### DIFF
--- a/kdumpctl
+++ b/kdumpctl
@@ -962,8 +962,24 @@ check_dump_feasibility()
 	check_kdump_feasibility
 }
 
+fadump_bootargs_append()
+{
+	if [[ -f "$FADUMP_APPEND_ARGS_SYS_NODE" ]]; then
+		local output=$( { echo "${FADUMP_COMMANDLINE_APPEND}" > "$FADUMP_APPEND_ARGS_SYS_NODE" ; } 2>&1)
+		if [ $? -eq 0 ]; then
+			output=$(cat "$FADUMP_APPEND_ARGS_SYS_NODE")
+			dinfo "fadump: additional parameters for capture kernel: '$output'"
+		else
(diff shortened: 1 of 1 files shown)


pfliu (approved):
LGTM

kdumpctl
```diff
@@ -962,8 +962,22 @@ check_dump_feasibility()
 	check_kdump_feasibility
 }
 
+fadump_bootargs_append()
+{
+	if [[ -f "$FADUMP_APPEND_ARGS_SYS_NODE" ]]; then
+		output=$( { echo "${FADUMP_COMMANDLINE_APPEND}" > "$FADUMP_APPEND_ARGS_SYS_NODE" ; } 2>&1)
```
prudo1, line 968: Hi Hari,
just a small nit. Could you please define `output` as `local`. Other than that the commit is fine.
Thanks
Philipp

kdumpctl
```diff
@@ -962,8 +962,22 @@ check_dump_feasibility()
...
+		if [ $? -eq 0 ]; then
+			output=$(cat "$FADUMP_APPEND_ARGS_SYS_NODE")
+			dinfo "fadump: additional parameters for capture kernel: '$output'"
+		else
+			dwarn "WARNING: failed to setup additional parameters for fadump capture kernel: '$output'"
+		fi
+	fi
+}
```
prudo1, line 976: noticed an other small nit. Could you please add a warning in case $FADUMP_APPEND_ARGS_SYS_NODE does not exist telling the user, that the current kernel does not support setting additional parameters.

hbathini: Thanks for the review. Updated the patch accordingly.

Issue #56, opened 2024-11-12
No such file or directory error in `kdump-utils.post` for`ppc64le`


gursewak1997:
When trying to build a Rawhide FCOS machine in ppc64le, it results in error

[2024-11-07T15:14:36.503Z] kdump-utils.post: No such file or directory
[2024-11-07T15:14:36.503Z] error: Installing packages: Running %post for kdump-utils: bwrap(/bin/sh): Child process killed by signal 2
[2024-11-07T15:14:36.503Z] failed to execute cmd-build: exit status 1
[2024-11-07T15:14:36.507Z] [Pipeline] }

Interestingly, this error is **only seen in ppc64le arch**.
It happened after the upgrade:
`kdump-utils-1.0.48-1.fc42.ppc64le`->`kdump-utils-1.0.49-1.fc42.ppc64le` (fedora-rawhide)

Changes in %post script in the update
https://src.fedoraproject.org/rpms/kdump-utils/c/c9c7de4907f308b9452f544f785af483d1b293b1?branch=rawhide

Reproducible: Always

Steps to Reproduce:
1. cosa fetch and cosa build in FCOS-Rawhide ppc64le
2. cosa build will fail with the above error

Clone of https://bugzilla.redhat.com/show_bug.cgi?id=2324486

pcdubs:
Also seeing this when building the ppc64le `fedora-bootc` images:

`error: Installing packages: Running %post for kdump-utils: bwrap(/bin/sh): Child process killed by signal 2`

Full logs:
https://kojipkgs.fedoraproject.org//work/tasks/8617/125778617/runroot.log

Koji task - https://koji.fedoraproject.org/koji/taskinfo?taskID=125778617

coiby:
Hi,

Thanks for reporting this bug.  Can you enable verbose mode when building FCOS or bootc images?
```sh
$ dnf4 --rpmverbosity=debug --verbose install kdump-utils
...
D: %post(kdump-utils-1.0.49-1.fc42.ppc64le): scriptlet start
D: %post(kdump-utils-1.0.49-1.fc42.ppc64le): execv(/bin/sh) pid 3908
D: Plugin: calling hook scriptlet_fork_post in selinux plugin
D: setexecfilecon: (/bin/sh, rpm_script_t) 
+ '[' 1 -eq 1 ']'
+ '[' -x /usr/lib/systemd/systemd-update-helper ']'
[...]

PR #58, opened 2024-11-22
Let %post scriptlet always exit successfully

coiby:
Resolves: https://github.com/rhkdump/kdump-utils/issues/56

Currently, kdump-utils fails to upgrade to
kdump-utils-1.0.48-1.fc42.ppc64le on ppc64le FCOS as servicelog_notify failed to be executed,

    # journalctl -t 'rpm-ostree(kdump-utils.post)'
    ... rpm-ostree(kdump-utils.post)[2376]: + touch /etc/kdump.conf
    ... rpm-ostree(kdump-utils.post)[2376]: + servicelog_notify --remove --command=/usr/lib/kdump/kdump-migrate-action.sh
    ... rpm-ostree(kdump-utils.post)[2376]: + servicelog_notify --add --command=/usr/lib/kdump/kdump-migrate-action.sh '--match=refcode="#MIGRATE" and serviceable=0' --type=EVENT --method=pairs_stdin
    ... rpm-ostree(kdump-utils.post)[2382]: No such file or directory

RPM scriplets should always return 0 status code to avoid breaking installs/upgrades/erases [1].

[1] https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/

Fixes: ea0b3afa ("Replace remaining occurrences of kexec-tools with kdump-utils")

### DIFF
--- a/kdump-utils.spec
+++ b/kdump-utils.spec
@@ -56,6 +56,10 @@ servicelog_notify --remove --command=/usr/lib/kdump/kdump-migrate-action.sh 2>/d
 servicelog_notify --add --command=/usr/lib/kdump/kdump-migrate-action.sh --match='refcode="#MIGRATE" and serviceable=0' --type=EVENT --method=pairs_stdin >/dev/null
 %endif
 
+# RPM scriptlet should always return 0. Otherwise it can break
+# installs/upgrades/erases.
+:
+
 
 %postun
 %systemd_postun_with_restart kdump.service


licliu (approved):
LGTM

cgwalters:
cc https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/1170

kdump-utils.spec
```diff
@@ -56,6 +56,10 @@ servicelog_notify --remove --command=/usr/lib/kdump/kdump-migrate-action.sh 2>/d
 servicelog_notify --add --command=/usr/lib/kdump/kdump-migrate-action.sh --match='refcode="#MIGRATE" and serviceable=0' --type=EVENT --method=pairs_stdin >/dev/null
 %endif
 
+# RPM scriptlet should always return 0. Otherwise it can break
```
cgwalters, line 59: No, I don't think scripts should always exit 0.

In image-based deployments `%post` scripts are run on a build server. I think the fundamental problem here is that the `%post` script is trying to do something interacting with ppc64le hardware (?) in `servicelog_notify`. That operation ultimately needs to be in. something like a systemd unit or triggered as part of kdump's existing system unit.

I think a better change would have been to just add `|| true` after the `servicelog_notify` in the short term.

But the reason RPM scripts should not *always* exit 0 is precisely because we *do* want to capture actual errors in them at build time! Let's say the script is generating some data that expects to be read.

If we get say an I/O error or a transient ENOSPC, we don't want to stumble on and generate a broken image.

What I really need to do is argue for a global switch in RPM that would make errors fatal by default that we can turn on for image builds.

coiby: Thanks for letting me know bootc prefers to not let RPM scriptlet always return 0 and suggesting to execute servicelog_notify elsewhere! Your reply also prompts me to learn about how bootc handles RPm script and help me realize the error should not be ignored. So I've filed https://issues.redhat.com/browse/RHEL-71502 to track this problem. 

Btw, I reached the conclusion that "RPM scriptlets should return 0" by reading the following texts in [Scriptlets :: Fedora Docs](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_syntax) which I think needs to be adapted for bootc, 

> All scriptlets MUST exit with the zero exit status. Because RPM in its default configuration does not execute shell scriptlets with the -e argument to the shell, excluding explicit exit calls (frowned upon with a non-zero argument!), the exit status of the last command in a scriptlet determines its exit status. Most commands in the snippets in this document have a “|| :” appended to them, which is a generic trick to force the zero exit status for those commands whether they worked or not. Usually the most important bit is to apply this to the last command executed in a scriptlet, or to add a separate command such as plain “:” or “exit 0” as the last one in a scriptlet. Note that depending on the case, other error checking/prevention measures may be more appropriate.

> Non-zero exit codes from scriptlets can break installs/upgrades/erases such that no further actions will be taken for that package in a transaction (see [Ordering](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#ordering)), which may for example prevent an old version of a package from being erased on upgrades, leaving behind duplicate rpmdb entries and possibly stale, unowned files on the filesystem. There are some cases where letting the transaction to proceed when some things in scriptlets failed may result in partially broken setup. It is however often limited to that package only whereas letting a transaction to proceed with some packages dropped out on the fly is more likely to result in broader system wide problems.

PR #65, opened 2025-01-06
99-kdump.conf: Omit nouveau module

licliu:
Resolves: https://issues.redhat.com/browse/RHEL-52304

The nouveau module provides no significant utility in second kernel, and it introduces firmware that occupies approximately 38MB of memory, which is critical in the constrained environment of kdump. Omit it helps reduce memory usage and optimize the crash recovery process.

### DIFF
--- a/99-kdump.conf
+++ b/99-kdump.conf
@@ -1,3 +1,4 @@
 dracutmodules=''
 add_dracutmodules=' kdumpbase '
 omit_dracutmodules=' rdma plymouth resume ifcfg earlykdump '
+omit_drivers+=' nouveau amdgpu '


daveyoung (changes requested):
Hi Lichen, thanks for the patch.  Sounds good to me as nouveau is known to be broken for kdump on many Lenovo hardwares (kdump kernel panic etc.)  It is not possible to have fixes in the near future.  

But the omit_dracutmodules should be omit_drivers instead.

BTW when you update the patch, it would be mention the Lenovo issues as well.

coiby (approved):

liutgnu (approved):
I tested this on my amdgpu machine, LGTM, so ack.

PR #69, opened 2025-01-21
Bring up the network before Open vSwitch bridge

coiby:
Resolves: https://issues.redhat.com/browse/RHEL-75547

Currently, kdump tries to bring up the same network as in 1st kernel as
we assume if the same network works in the 1st kernel, there is no
reason it won't work in 1st kernel.

But for Open vSwitch (OVS) bridge, there are two challenges,
1. Extra memory requirement
   For example, crashkernel=260M will be needed for OVS bridge while
   only crashkernel=192M is sufficient for default network for F40 cloud
   image.

2. A bit complex dependencies
   An extra dracut module is needed to install dependent executables,
   helper scripts, configuration files and a NetworkManager plugin.

To avoid the above challenges, we try to bring up the network before OVS
bridge instead. OVS bridge is set up by ovs-configuration.service
[1][2], we can assume the previous network has worked.

When ovs-configuration.service is started, it will bring down the
original connection profiles. So we should go through the inactive
connection profiles to find out what connection profiles is for bringing
up the physical network interface.

```
How to test this patch with the help of configure-ovs.sh?
=========================================================

1. Extract configure-ovs.sh from [2]

2. Install necessary packages for configure-ovs.sh
    dnf install openvswitch -yq
    dnf install NetworkManager-ovs nmap-ncat -yq

    systemctl enable --now openvswitch

    # restart NM so the ovs plugin can be activated
    systemctl restart NetworkManager

3. Assume the network interface used for creating an Ovs bridge is
   "ens2", use configure-ovs.sh to create an Ovs bridge,

    interface=ens2
    mkdir -p /etc/ovnk
    echo $interface > /etc/ovnk/iface_default_hint
    bash configure-ovs.sh OVNKubernetes

4. (Optional) If you want to make the created Ovs bridge survive a
   reboot, simply make the created NM connections created by
   configure-ovs.sh persist,

    cp /run/NetworkManager/system-connections/* /etc/NetworkManager/system-connections/

If you need to create an Ovs bridge on top of a bonding network, use the
following commands for step 3,

    nmcli con add type bond ifname bond0
    nmcli con add type ethernet ifname eth0 master bond0
    nmcli con add type ethernet ifname eth1 master bond0

    echo bond0 > /etc/ovnk/iface_default_hint
    bash configure-ovs.sh OVNKubernetes
````

Note
1. For RHEL9, openvswitch package needs to installed from another repo,
    cat << 'EOF' > /etc/yum.repos.d/ovs.repo
    [rhosp-rhel-9-fdp-cdn]
    name=Red Hat Enterprise Linux Fast Datapath $releasever - $basearch cdn
    baseurl=http://rhsm-pulp.corp.redhat.com/content/dist/layered/rhel9/$basearch/fast-datapath/os/
    enabled=1
    gpgcheck=0
    EOF

    dnf install openvswitch3.3 -yq

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/units/ovs-configuration.service.yaml
[2] https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml

### DIFF
--- a/dracut/99kdumpbase/module-setup.sh
+++ b/dracut/99kdumpbase/module-setup.sh
@@ -488,24 +488,6 @@ kdump_setup_vlan() {
     _save_kdump_netifs "$_parent_netif"
 }
 
-kdump_setup_ovs() {
-    local _netdev="$1"
-    local _dev _phy_if
-
-    _phy_if=$(ovs_find_phy_if "$_netdev")
-
-    if kdump_is_bridge "$_phy_if"; then
-        kdump_setup_vlan "$_phy_if"
-    elif kdump_is_bond "$_phy_if"; then
-        kdump_setup_bond "$_phy_if" || return 1
-    elif kdump_is_team "$_phy_if"; then
-        derror "Ovs bridge over team is not supported!"
-        exit 1
-    fi
-
-    _save_kdump_netifs "$_phy_if"
-}
-
 # setup s390 znet
 kdump_setup_znet() {
     local _netif
@@ -542,28 +524,6 @@ kdump_get_remote_ip() {
     echo "$_remote"
 }
 
-# Find the physical interface of Open vSwitch (Ovs) bridge
-#
-# The physical network interface has the same MAC address as the Ovs bridge
-ovs_find_phy_if() {
-    local _mac _dev
-    _mac=$(kdump_get_mac_addr "$1")
-
-    for _dev in $(ovs-vsctl list-ifaces "$1"); do
-        if [[ $_mac == $(< /sys/class/net/"$_dev"/address) ]]; then
-            echo -n "$_dev"
-            return
-        fi
-    done
-
-    return 1
-}
-
-# Tell if a network interface is an Open vSwitch (Ovs) bridge
-kdump_is_ovs_bridge() {
-    [[ $(_get_nic_driver "$1") == openvswitch ]]
-}
-
 # Collect netifs needed by kdump
 # $1: destination host
 kdump_collect_netif_usage() {
@@ -587,9 +547,6 @@ kdump_collect_netif_usage() {
         kdump_setup_team "$_netdev"
     elif kdump_is_vlan "$_netdev"; then
         kdump_setup_vlan "$_netdev"
-    elif kdump_is_ovs_bridge "$_netdev"; then
-        has_ovs_bridge=yes
-        kdump_setup_ovs "$_netdev"
     fi
     _save_kdump_netifs "$_netdev"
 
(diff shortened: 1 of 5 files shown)


igsilya:
> OVS bridge is set up by ovs-configuration.service
> [1][2], we can assume the previous network has worked.
>
> When ovs-configuration.service is started, it will bring down the
> original connection profiles. So we should go through the inactive
> connection profiles to find out what connection profiles is for bringing
> up the physical network interface.

This is very specific to OpenShift.  Is there a specific reason for tying up the solution to OpenShift?  Seems a bit strange.

coiby:
> > OVS bridge is set up by ovs-configuration.service
> > [1][2], we can assume the previous network has worked.
> > When ovs-configuration.service is started, it will bring down the
> > original connection profiles. So we should go through the inactive
> > connection profiles to find out what connection profiles is for bringing
> > up the physical network interface.
> 
> This is very specific to OpenShift. Is there a specific reason for tying up the solution to OpenShift? Seems a bit strange.

Hi Ilya, thanks for your feedback! Currently, all use cases I see is about OpenShift + OVS. If there are other use cases or the way to set up OVS bridge by ovs-configuration.service is not general enough, can you point me to those use cases? Thanks!

coiby:
> > Hi Ilya, thanks for your feedback! Currently, all use cases I see is about OpenShift + OVS. If there are other use cases or the way to set up OVS bridge by ovs-configuration.service is not general enough, can you point me to those use cases? Thanks!
> 
> There is the OSP use case. @christophefontaine can you shine some light on how it configures OVS?

@chaudron Thanks for bringing up the OSP use case. It seems OCP uses [os-net-config](https://docs.openstack.org/os-net-config/latest/) to set up OVS.  After playing with it for a while,  I find os-net-config doesn't depend on an active network interface. So the assumption in this PR that an existing connection will brought down to establish OVS bridge in the patch may not necessarily apply to OSP. Then the approach adopted by this PR may not be as robust as the previous approach which adds OVS support in kdump initramfs. 

On the other hand, adding OVS support in kdump intramfs will demand more memory for kdump kernel and the current default crashkernel value won't work for systems with RAM < 64 GB. Do you know roughly what percentage of systems use RAM smaller than 64GB?

chaudron:
> After spending some time digging out how OSP sets up OVS bridge, I think it's better to only support OCP for for now. 

Thanks for your analysis! I'll let @christophefontaine respond to whatever they need OSP support or not.

licliu (approved):
LGTM

PR #70, opened 2025-01-21
tests: set up custom mirror before essential-requires

coiby:
By default, the embedded task essential-requires runs before all prepare tasks,
        queued prepare task #1: essential-requires on client and server
        queued prepare task #2: Set root password on client and server
        queued prepare task #3: Use custom mirror on client and server
        queued prepare task #4: Install built RPM on client

And it takes 13m to finish the SSH dumping tests. By running task "Use custom mirror" early, the test could be finished within 3m.

### DIFF
--- a/tests/plans/main.fmf
+++ b/tests/plans/main.fmf
@@ -9,17 +9,19 @@ provision:
 
 
 prepare:
+  - name: Use custom mirror
+    # Run this task before task essential-requires
+    order: 30
+    how: shell
+    script:
+      - test -v CUSTOM_MIRROR && sed -e 's/^metalink=/#metalink=/g' -e "s|^#baseurl=http://download.example/pub/fedora/linux|baseurl=${CUSTOM_MIRROR}|g" -i.bak /etc/yum.repos.d/fedora*.repo || true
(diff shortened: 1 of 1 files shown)


licliu (approved):
Hi @coiby, this change looks good to me, but a small question about the CUSTOM_MIRROR, I find the only reference here.
https://github.com/rhkdump/kdump-utils/blob/27df3385b059693e983190c8ad80bd1f7e12db6c/tools/run-integration-tests.sh#L30
I think it is a China mirror, I don't know if it can be accessed normally in other countries and regions.

coiby:
> Hi @coiby, this change looks good to me, 

Hi @licliu, thanks for reviewing the patch! 

> but a small question about the CUSTOM_MIRROR, I find the only reference here.
> https://github.com/rhkdump/kdump-utils/blob/27df3385b059693e983190c8ad80bd1f7e12db6c/tools/run-integration-tests.sh#L30 
> 
> I think it is a China mirror, I don't know if it can be accessed normally in other countries and regions.
Yes, China mirror is used because the integration tests run on machines located in China. But the mirror is also accessible to other countries because it's public.

PR #73, opened 2025-02-25
Enable tests against rawhide

coiby:
Resolves: https://issues.redhat.com/browse/KEX-203

By running kdump-utils tests against Fedora rawhide, we can catch regressions earlier. This will also cover latest dracut as kdump heavily depends on dracut.


licliu (approved):
LGTM

PR #79, opened 2025-03-21
s/kexec-tools/kdump-utils

coiby:
Replace kexec-tools with kdump-utils when appropriate.

Note zanata-notes.txt and sources have been deleted as they are invalid now.


licliu (approved):
LGTM

PR #81, opened 2025-04-24
mkdumprd: include tpm2-tss dracut module for encrypted storages

vittyvk:
LUKS volumes can be automatically unlocked by TPM devices and this comes handy for kdump as there's no need to a manual action. This, however, doesn't work by default:

 systemd-cryptsetup[494]: Automatically discovered security TPM2 token unlocks volume.
 systemd-cryptsetup[494]: Could not find TPM2 device: Operation not supported
 systemd-cryptsetup[494]: TPM2 operation failed, falling back to traditional unlocking: Operation not supported

The problem is that we need to have libcryptsetup-token-systemd-tpm2.so and the corresponding tools/libraries in the initramfs and this is done by 'tpm2-tss' module which is not in the default set. Include it when 'tpm2' tool (tpm2-tools package) is present.

In theory, we can be smarter and try to detect 'systemd-tpm2' tokens with cryptsetup:

 # cryptsetup luksDump /dev/vda2
 LUKS header information
 Version:       	2
 Epoch:         	394
 ...

 Tokens:
   1: systemd-tpm2
 	tpm2-hash-pcrs:   7
 	tpm2-pcr-bank:    sha256

but the risk is that tokens can change after kdump initramfs is created. For now, stick to a simpler but more robust solution.

### DIFF
--- a/mkdumprd
+++ b/mkdumprd
@@ -330,6 +330,10 @@ check_crypt()
 
 if ! check_crypt; then
 	dwarn "Warning: Encrypted device is in dump path, which is not recommended, see kexec-kdump-howto.txt for more details."
+	# TPM device can be used to automatically unlock encrypted volumes
+	if has_command tpm2; then
+		dracut_args+=(--add "tpm2-tss")
+	fi
 fi
 
 # firstly get right SSH_KEY_LOCATION


coiby:
Hi @vittyvk ,

Do you have an agenda to support this case in RHEL? If we can wait,  I'll prefer a more general solution [[PATCH v9 0/8] Support kdump with LUKS encryption by reusing LUKS volume keys](https://patchew.org/linux/20250502011246.99238-1-coxu@redhat.com/20250502011246.99238-7-coxu@redhat.com/). Currently the patch set has been pulled into Andrew Morton's nm-everything tree and there is a good chance it will be pulled into Linus's tree.

vittyvk:
Hey @coiby , yes, 'reusing LUKS keys' is a much more interesting feature indeed :-) Here, however, we may be solving one additional use-case: imagine we want to dump to a LUKS device which is not actually used in the main system (e.g. a special dedicated device for dumps) which can be unlocked with TPM. Then, we can't pass LUKS key to it from the main kernel as it doesn't have it. I'm not sure if it is so important though.

coiby:
Thanks for proposing another scenario! Unfortunately, kdumpctl needs to unlock the device to check if disk space will be available for vmcore dumping. However it doesn't know how to unlock a LUKS device. If we want to support this additional case, we will need to add these code as well. Otherwise, kdumpctl will fail to build the initramfs,
```
kdump: Rebuilding /boot/initramfs-6.14.0-rc1+kdump.img
kdump: Dump target "UUID=a3a60e7b-d671-4629-bc5c-b809bbcbde15" is neither mounted nor configured as "noauto"
kdump: mkdumprd: failed to make kdump initrd
```

Btw, my work to support kdump with LUKS encryption by reusing LUKS volume keys [has been merged into Linus
s tree](https://github.com/torvalds/linux/commit/180cf31af7c313790f1e0fba1c7aa42512144dd5). While doing the corresponding user space part in kdump-utils and anaconda, I'll also evaluate the necessity of this PR.

PR #85, opened 2025-05-19
RFC: Send standard error of capture service to TTY

sourabhjains:
On PowerPC, makedumpfile fails to print the ETA during dump collection if the StandardError of the kdump capture service is configured to journal+console. Changing it to tty fixes the issue.

Changing StandardError to tty may not be the right approach, but this patch is being sent to discuss possible solutions to the problem, since the same service is used on other architectures as well.

In the kdump kernel, the system does have /dev/console, but for some reason, makedumpfile running via the capture service is not able to print the ETA during dump collection.

Running the following command:
  ```$ echo "hello" > /dev/console```

prints hello on the console in the kdump emergency shell.

## Summary by Sourcery

Bug Fixes:
- Configure kdump-capture.service to send stderr to tty to restore makedumpfile’s ETA output during dump collection on PowerPC

### DIFF
--- a/dracut/99kdumpbase/kdump.sh
+++ b/dracut/99kdumpbase/kdump.sh
@@ -182,7 +182,7 @@ dump_fs() {
     KDUMP_LOG_DEST=$_dump_fs_path/
     KDUMP_LOG_OP="mv '$KDUMP_LOG_FILE' '$KDUMP_LOG_DEST/'"
 
-    $CORE_COLLECTOR /proc/vmcore "$_dump_fs_path/vmcore-incomplete"
+    $CORE_COLLECTOR /proc/vmcore "$_dump_fs_path/vmcore-incomplete" > /dev/console 2>&1
     _dump_exitcode=$?
     if [ $_dump_exitcode -eq 0 ]; then
         sync -f "$_dump_fs_path/vmcore-incomplete"


coiby:
@sourabhjains Great finding! Since this issue is arch-independent and can easily be reproduced, I think we should report it to systemd developers together with your reproducer. If they don't think it's a systemd bug, we can manually copy  stder to /dev/console manually when invoking makedumpfile in kdump.sh

sourabhjains:
@coiby It's been a month, and yet there's no comment on the issue I reported regarding the systemd service not showing ETA updates on the console.

I verified that forwarding makedumpfile's stderr to /dev/console in the kdump capture script solves the problem, as you suggested https://github.com/rhkdump/kdump-utils/pull/85#issuecomment-3003343475

Shall I send an updated patch with the change in the kdump script to manually copy stderr to /dev/console when invoking makedumpfile?

Please share your opinion.

coiby:
Thanks for the update!  I also tried digging into the systemd code but can't find an easy way. So yes, please send an updated patch to manually copy stderr to /dev/console when invoking makedumpfile which shall be much easier way.

sourabhjains:
Any feedback on this pull request.

coiby (approved):

PR #88, opened 2025-05-30
Fix shfmt and shellcheck warnings

coiby:
shfmt and shellcheck have found some issues in the kdump-utils code. Although current majority of the issues there are real problems, it's good to fix them so news issues can be caught more easily.

## Summary by Sourcery

Enforce consistent shell formatting and resolve ShellCheck warnings across kdump-utils by reformatting scripts, adding missing quotes, and adjusting CI and build dependencies.

Enhancements:
- Reformat all shell scripts using shfmt to standardize indentation, function definitions, and bracket spacing
- Quote variable expansions and command substitutions throughout code and tests to prevent word-splitting and ShellCheck errors
- Insert shellcheck disable pragmas and source annotations where necessary to suppress false positives
- Update test and spec files to properly quote parameters and accommodate shellspec directives

Build:
- Add gawk as an RPM build dependency in the spec file

CI:
- Install gawk in CI runner and refine shellcheck invocation to include spec files
- Disable specific ShellCheck checks in CI for unit test patterns

### DIFF
--- a/dracut/99kdumpbase/module-setup.sh
+++ b/dracut/99kdumpbase/module-setup.sh
@@ -11,7 +11,10 @@ _save_kdump_netifs() {
 }
 
 _get_kdump_netifs() {
-    echo -n "${!unique_netifs[@]} ${!ovs_unique_netifs[@]}"
+    local -a _all_netifs
+
+    _all_netifs=("${!unique_netifs[@]}" "${!ovs_unique_netifs[@]}")
+    echo -n "${_all_netifs[@]}"
 }
 
 kdump_module_init() {


kdump-dep-generator.sh
```diff
@@ -11,13 +11,13 @@
...
 fi
 
 systemd_dir=/usr/lib/systemd/system
 kdump_wants=$dest_dir/kdump.service.wants
 
 if is_ssh_dump_target; then
-    mkdir -p $kdump_wants
-    ln -sf $systemd_dir/network-online.target $kdump_wants/
+	mkdir -p "$kdump_wants"
+	ln -sf $systemd_dir/network-online.target "$kdump_wants"/
```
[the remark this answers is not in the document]
licliu: +1

kdump-lib-initramfs.sh
```diff
@@ -4,10 +4,12 @@
 # not be the default shell. Any code added must be POSIX compliant.
 
 DEFAULT_PATH="/var/crash/"
+# shellcheck disable=SC2034
 DEFAULT_SSHKEY="/root/.ssh/kdump_id_rsa"
 KDUMP_CONFIG_FILE="/etc/kdump.conf"
-FENCE_KDUMP_CONFIG_FILE="/etc/sysconfig/fence_kdump"
```
licliu, line 9: The default value of FENCE_KDUMP_CONFIG_FILE was removed instead of being moved to kdump-lib.sh.

kdump-lib.sh
```diff
@@ -764,7 +768,7 @@ prepare_cmdline()
 	out+="$append "
 
 	id=$(get_bootcpu_apicid)
-	if [[ -n "${id}" ]]; then
+	if [[ -n ${id} ]]; then
```
licliu, line 771: Is there any difference between using `$id` and `${id}` here?

kdump-lib-initramfs.sh
coiby: Good catch! Fixed in new version.

coiby:
Hi @licliu,

Thanks for carefully reviewing the PR! I've pushed a new version which also 1) rebases onto latest main branch and 2) fixed [an issue caught by Gemini early](https://github.com/coiby/kdump-utils/pull/14#discussion_r2122402079) but I forgot to fix besides addressing your feedback ([diff change](https://github.com/rhkdump/kdump-utils/compare/7fa77f8705a80e2112417e397dcfd4181d1165f5..04dd80332e61001f32b1fe43f3ec9bd16d1daee2)) .

licliu (approved):
Thanks @coiby, this version looks good to me.

PR #89, opened 2025-06-04
mkdumprd: mount the root partition when FIPS is enabled with no boot partition

coiby:
Resolves: https://issues.redhat.com/browse/RHEL-83641

Currently, remote vmcore dumping fails when FIPS mode is enabled on a
system with no boot partition,

    [    3.440880] systemd[1]: Starting dracut-pre-pivot.service - dracut pre-pivot and cleanup hook...
    ...
    [    3.461906] dracut-pre-pivot[627]: fips-noboot: start
    [    3.638070] dracut: FATAL: You have to specify boot=<boot device> as a boot option for fips=1
    [    3.639323] dracut: Refusing to continue
    ...
    [    3.475819] systemd[1]: Poweroff requested from client PID 649 ('systemctl') (unit dracut-pre-pivot.service)...
    [    3.477611] systemd[1]: Shutting down.
    ...
    [    3.914537] systemd-shutdown[1]: Powering off.
    [    3.916464] ACPI: PM: Preparing to enter system sleep state S5
    [    3.917614] reboot: Power down

This happens because the fips dracut module shutdowns the system when it
can't find /sysroot/boot. So mount the root partition to /sysroot to fix
this issue.

## Summary by Sourcery

Fix remote vmcore dumping failure on FIPS-enabled systems without a separate boot partition by mounting the root partition to /sysroot in mkdumprd.

Bug Fixes:
- Mount the root partition to /sysroot when FIPS mode is enabled without a boot partition to prevent dracut from refusing to continue.

Tests:
- Add a spec test to verify get_kdump_mntpoint_from_target returns "/sysroot" for root partitions.

### DIFF
--- a/mkdumprd
+++ b/mkdumprd
@@ -433,7 +433,7 @@ if ! is_fadump_capable; then
 	# If /boot is not on a separate partition, the fips dracut module will
 	# link /sysroot/boot to /boot. So we need to mount the root partition
 	# to /sysroot beforehand.
-	if [[ $(cat /proc/sys/crypto/fips_enabled) == 1  ]]; then
+	if [[ $(cat /proc/sys/crypto/fips_enabled 2> /dev/null) == 1  ]]; then
 		_boot_source=$(findmnt -n -o SOURCE --target /boot)
 		if mountpoint -q /boot; then
 			dracut_args+=(--add-device "$_boot_source")


licliu (approved):
LGTM

PR #91, opened 2025-06-24
reset-crashkernel for multiple grub entries.

licliu:
When there are multiple grub entries for the same kernel, get_grub_kernel_boot_parameter()
uses the first matching $_para value, which may cause inconsistent behavior in the function
that calls it.

This PR tries to fix https://issues.redhat.com/browse/RHEL-79964 by checking all grub enties of the given kernel.
All matched grub entries will be updated if possible.

## Summary by Sourcery

Update kdumpctl to handle multiple GRUB entries for the same kernel by gathering and applying boot parameters across all matches instead of only the first occurrence.

Bug Fixes:
- Ensure consistent retrieval of crashkernel boot parameters across multiple GRUB entries

Enhancements:
- Introduce get_all_grub_kernel_boot_parameter function to collect and deduplicate boot parameters across all matching GRUB entries
- Update find_all_kernels to use the new multi-entry parameter collection and iterate through all menuentries when no kernel is specified

### DIFF
--- a/kdumpctl
+++ b/kdumpctl
@@ -1596,7 +1596,9 @@ reset_crashkernel()
 					fi
 				fi
 			else
-				_dump_mode="$(get_dump_mode_by_kernel "$_kernel")"
+				if ! _dump_mode="$(get_dump_mode_by_kernel "$_kernel")"; then
+					exit 1
+				fi
 				_new_ck=$(kdump_get_arch_recommend_crashkernel "$_dump_mode")
 			fi
 		else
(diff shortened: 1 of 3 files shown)


coiby:
Hi @licliu,

For me, it's easier to reason about the reset_crashkernel function if we change get_grub_kernel_boot_parameter to query kernel parameter by index instead of by kernel path because ` kdumpctl reset-crashkernel --kernel=ALL` can also support reset crashkernel for multiple bootloader entries. If you feel the same way, can you use this approach instead?

And can you also add a unit test to catch the issue that this PR tries to resolve?

licliu:
@coiby Thanks! I will do that next version.

licliu:
fix issues found by Gemini.

coiby:
Hi @licliu ,

Thanks for adopting a new approach which makes it much easier to reason about the code! And also thanks for adding more unit test cases! The PR LGTM! One trivial issue is the new unit test cases have some format issues. I've used [shellspec/altshfmt](https://github.com/shellspec/altshfmt) to reformat the code.

PR #92, opened 2025-06-26
Maximize the reserved size in osbuild case

pfliu:
At present, the deduction of the proper crashkernel value depends on
    detecting platform details. However, more and more platforms are now
    deployed using OSTree images, which means kdump-utils cannot retrieve
    complete platform information. This often leads to an underestimation of
    the memory required by the kdump kernel, increasing the risk of
    out-of-memory (OOM) issues.
    
    To mitigate this situation, we choose to maximize the reserved
    crashkernel size in the osbuild case, and recommend users to update the
    kernel command line and release the excess reserved memory after reboot.
    
    That is done by: kdumpctl stop; echo $_recommend_value >/sys/kernel/kexec_crash_size; kdumpctl start

### DIFF
--- a/kdump-lib.sh
+++ b/kdump-lib.sh
@@ -11,6 +11,7 @@ fi
 FADUMP_ENABLED_SYS_NODE="/sys/kernel/fadump/enabled"
 FADUMP_REGISTER_SYS_NODE="/sys/kernel/fadump/registered"
 FADUMP_APPEND_ARGS_SYS_NODE="/sys/kernel/fadump/bootargs_append"
+maximize_crashkernel=0
 
 is_uki()
 {
@@ -41,7 +42,7 @@ is_aws_aarch64()
 
 is_sme_or_sev_active()
 {
-	journalctl -q --dmesg --grep "^Memory Encryption Features active: AMD (SME|SEV)$" >/dev/null 2>&1
+	$maximize_crashkernel || journalctl -q --dmesg --grep "^Memory Encryption Features active: AMD (SME|SEV)$" >/dev/null 2>&1
 }
 
 has_command()
@@ -846,12 +847,18 @@ get_recommend_size()
 
 has_mlx5()
 {
-	[[ -d /sys/bus/pci/drivers/mlx5_core ]]
+	$maximize_crashkernel || [[ -d /sys/bus/pci/drivers/mlx5_core ]]
 }
 
 has_aarch64_smmu()
 {
-	ls /sys/devices/platform/arm-smmu-* 1> /dev/null 2>&1
+	$maximize_crashkernel || ls /sys/devices/platform/arm-smmu-* 1> /dev/null 2>&1
+}
+
+is_aarch64_64k_kernel()
+{
+	local _kernel="$1"
+	$maximize_crashkernel || echo "$_kernel" | grep -q 64k
 }
 
 is_memsize() { [[ "$1" =~ ^[+-]?[0-9]+[KkMmGgTtPbEe]?$ ]]; }
@@ -995,6 +1002,9 @@ kdump_get_arch_recommend_crashkernel()
 	local _arch _ck_cmdline _dump_mode
 	local _delta=0
 
+	# osbuild deploies rpm on chroot environment. kdump-utils has no opportunity
+	# to deduce the exact memory cost on the real target.
+	maximize_crashkernel=$(is_ostree)
 	if [[ -z "$1" ]]; then
 		if is_fadump_capable; then
 			_dump_mode=fadump
@@ -1022,7 +1032,7 @@ kdump_get_arch_recommend_crashkernel()
 		fi
 
 		# the naming convention of 64k variant suffixes with +64k, e.g. "vmlinuz-5.14.0-312.el9.aarch64+64k"
-		if echo "$_running_kernel" | grep -q 64k; then
+		if is_aarch64_64k_kernel "$_running_kernel"; then
 			# Without smmu, the diff of MemFree between 4K and 64K measured on a high end aarch64 machine is 82M.
 			# Picking up 100M to cover this diff. And finally, we have "2G-4G:356M;4G-64G:420M;64G-:676M"
 			((_delta += 100))
--- a/kdumpctl
+++ b/kdumpctl
@@ -1051,6 +1051,83 @@ check_final_action_config()
 	esac
 }
 
+get_recommend_crashkernel_value()
+{
+	local _crashkernel_str=$1
+	local _total_mem_kib
+	local _total_mem
+
+	_total_mem_kib=$(grep MemTotal /proc/meminfo | awk '{print $2}')
+	_total_mem=$((_total_mem_kib * 1024))
+
+	# Main loop
+	IFS=',' read -ra entries <<< "$_crashkernel_str"
+	for entry in "${entries[@]}"; do
+		IFS=':' read -r range size <<< "$entry"
+		# Parse range
+		if [[ $range == *-* ]]; then
+			IFS='-' read -r lower upper <<< "$range"
+			lower_bytes=$(to_bytes "$lower")
+			if [[ -z "$upper" ]]; then
+				upper_bytes=$(to_bytes 10T)  # for open-ended range (e.g., 2G-)
+			else
+				upper_bytes=$(to_bytes "$upper")
+			fi
+			if (( _total_mem >= lower_bytes && _total_mem < upper_bytes )); then
+				echo "$size"
+				return
+			fi
+		fi
+	done
+}
+
+get_crashkernel_low_value()
+{
+	mapfile -t lines < <(grep -n "Crash kernel" /proc/iomem)
+
+	if [ ${#lines[@]} -eq 1 ]; then
+		return 0
+	fi
+
+	first_line=$(grep "Crash kernel" /proc/iomem | sed -n '1p')
+	# Extract start and end addresses
+	start_hex=$(echo "$first_line" | awk '{split($1, a, "-"); print a[1]}')
+	end_hex=$(echo "$first_line" | awk '{split($1, a, "-"); print a[2]}')
+	# Convert hex to decimal
+	start_dec=$((0x$start_hex))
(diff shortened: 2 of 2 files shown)


pfliu:
I think even for local dump, if the OSTree image, containing both 4K and 64K variant, is built on aarch64-4K platform and finally the user deploys the 64K variant, we will face the OOM issue

kdump-lib.sh
```diff
@@ -995,6 +1002,9 @@ kdump_get_arch_recommend_crashkernel()
 	local _arch _ck_cmdline _dump_mode
 	local _delta=0
 
+	# osbuild deploies rpm on chroot environment. kdump-utils has no opportunity
+	# to deduce the exact memory cost on the real target.
```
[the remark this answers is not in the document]
pfliu: resolved in the forced update version

kdumpctl
```diff
@@ -1051,6 +1051,83 @@ check_final_action_config()
 	esac
 }
 
+get_recommend_crashkernel_value()
+{
```
coiby, line 1055: I think `get_recommend_crashkernel_value`  is duplicating `kdump_get_arch_recommend_size`. Btw, there is also a problem with `get_recommend_crashkernel_value` i.e it uses MemTotal from /proc/meminfo which is actually smaller than the system RAM.

kdump-lib.sh
```diff
@@ -995,6 +1002,9 @@ kdump_get_arch_recommend_crashkernel()
 	local _arch _ck_cmdline _dump_mode
 	local _delta=0
 
+	# osbuild deploys rpm on chroot environment. kdump-utils has no opportunity
+	# to deduce the exact memory cost on the real target.
+	maximize_crashkernel=$(is_ostree)
```
coiby, line 1007: osbuild is different from ostree. Osbuild can be used to build  normal guest image or ostree image.

kdumpctl
```diff
@@ -1051,6 +1051,83 @@ check_final_action_config()
...
+	_crashkernel=$(sed -n 's/.*crashkernel=\([^ ]*\).*/\1/p' /proc/cmdline)
+	[[ $(kdump_get_conf_val auto_reset_crashkernel) == no ]] && return
+
+	if ! echo "$_crashkernel" | grep -q -E "(fadump|fadump=0)"; then
+		_recommend=$(kdump_get_arch_recommend_crashkernel "kdump")
+		_recommend_value=$(get_recommend_crashkernel_value "$_recommend")
+		_low_mem=$(get_crashkernel_low_value)
+		# The crashkernel formula does not include the crashkernel low value
+		# but the kexec_crash_size does.
+		_recommend_value=$(memsize_add "$_recommend_value" "$_low_mem")
```
coiby, line 1120: According to https://docs.kernel.org/arch/arm64/kdump.html#crashkernel-size, for crashkernel=size, the kernel searches the low memory area by default. So crashkernel does include the crashkernel low value. So I think the logic here doesn't make sense.

coiby (changes requested):

kdumpctl
pfliu: I will remove it and resort to kdump_get_arch_recommend_size(). Thanks

kdump-lib.sh
pfliu: I think I need is_osbuild() here. And I will figure out a way to implement it. Thanks

kdumpctl
pfliu: The system's behavior diverges. If low memory is enough, there will no extra crashkernel,low is allocated. This is the case you mentioned here. But if low memory is not enough, the crashkernel=size specifies the volume of the memory reserved in high memory, not including the extra 128M crashkernel,low

pfliu: In the original solution, if there is no extra crashkernel.low is reserved, get_crashkernel_low_value() will 'echo 0', which means zero volume is reserved. And it can cover the 1st case

kdumpctl
```diff
@@ -1051,6 +1051,83 @@ check_final_action_config()
...
+	# Extract start and end addresses
+	start_hex=$(echo "$first_line" | awk '{split($1, a, "-"); print a[1]}')
+	end_hex=$(echo "$first_line" | awk '{split($1, a, "-"); print a[2]}')
+	# Convert hex to decimal
+	start_dec=$((0x$start_hex))
+	end_dec=$((0x$end_hex))
+	# Calculate size in bytes
+	size_bytes=$((end_dec - start_dec + 1))
+	echo "$size_bytes"
+}
```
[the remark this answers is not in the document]
pfliu: Adopt this suggestion

pfliu:
@coiby, I have updated patches according to your comments. It should resolve all issues except the crashkernel low, which I think it is correct.
When you have time, could you help to review it again?
Thanks

kdumpctl
coiby: Thanks for the explanation! Combined with a few experiments done on crashkernel,  now I see what you mean. But `get_crashkernel_low_value` and the comment "The crashkernel formula does not include the crashkernel low value" are still confusing. Maybe you can change the function name to improve the code readability.

Btw, I'm curious to ask when will the case where low memory is not enough happen? For aarch64, in my experiment, even crashkernel=2G won't use the high memory,
```
[root@rhel10 ~]# kdumpctl showmem
kdump: Reserved 2048MB memory for crash kernel
[root@rhel10 ~]# cat /proc/cmdline 
root=UUID=3e69ec88-a92a-4de9-a000-e394a3f7d3ea console=tty0 console=ttyS0,115200n8 no_timer_check systemd.firstboot=off crashkernel=2G
[root@rhel10 ~]# grep Crash /proc/iomem 
  80000000-ffffffff : Crash kernel
```  

Another question is I don't understand why `crashkernel=512M,low`  fails but `crashkernel=512M` works or is it a known bug?
```
[root@rhel10 ~]# uname -r
6.12.0-30.el10.aarch64


[root@rhel10 ~]# grep Crash /proc/iomem 
[root@rhel10 ~]# cat /proc/cmdline 
root=UUID=3e69ec88-a92a-4de9-a000-e394a3f7d3ea console=tty0 console=ttyS0,115200n8 no_timer_check systemd.firstboot=off crashkernel=512M,low
[root@rhel10 ~]# kdumpctl showmem
kdump: Reserved 0MB memory for crash kernel

[root@rhel10 ~]# cat /proc/cmdline 
root=UUID=3e69ec88-a92a-4de9-a000-e394a3f7d3ea console=tty0 console=ttyS0,115200n8 no_timer_check systemd.firstboot=off crashkernel=512M
[root@rhel10 ~]# grep Crash /proc/iomem 
  e0000000-ffffffff : Crash kernel
[root@rhel10 ~]# kdumpctl showmem
kdump: Reserved 512MB memory for crash kernel
```

coiby (changes requested):

kdumpctl
coiby: I have another question. I have impression that the default crashkernel value specify the total memory required for crash kernel. Let's say crashkernel=2G-4G:256M,4G-64G:320M,64G-:576M and system RAM is larger than 64G. When low memory is not enough, the memory reserved will become `576M + 128M`. Does the system really need an extra `128M` or  is `576M` sufficient? If 576M is sufficient, then there is no need to for `get_crashkernel_low_value` and related code.

kdump-lib.sh
```diff
@@ -273,6 +274,20 @@ is_ostree()
 	test -f /run/ostree-booted
 }
 
+# Decide whether it runs in sandbox or not
+is_osbuild()
+{
```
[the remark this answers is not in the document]
pfliu: Oh, thanks, I will use it

kdumpctl
coiby: Hi @pfliu I have a third question which somehow isn't show together with other comments.

coiby:
> Hi Coiby
> 
> For the first question: where low memory is not enough happen? Before crashkernel= is handled by the kernel, there are several subsystem has been initialized. If they occupy too many memory below 4G, there will be no enough memory below 4G left for crashkernel=. Also, there are some machines with sparse physical memory below 4G. It is short of memory below 4GB.

Thanks for your answer! 

> 
> For the second question why `crashkernel=512M,low` fails but `crashkernel=512M` works or is it a known bug? It is not a bug. crashkernel=512M,low will only try its luck under 4G. On the other hand, crashkernel=512M will try it luck under 4G in the first place, if it fails, it will try its luck from memory address greater than 4G. In the latter case, it only reserves additional 128MB below 4GB for bounce-DMA.

But if you check my pasted logs, for `crashkernel=512M`, the allocated memory for crash kernel is `e0000000-ffffffff` which belongs to the low memory. Besides only one line matches "Crash kernel" so there is no additional 128MB below 4GB.

> 
> I will improve the notes around the get_crashkernel_low_value().
> 
> Besides that, I just realize that this solution should skip the 2G-4G:256M. Because in this case, it usually a guest without smmu or mlx netcard. If we reserve large chunk of memory (1G) for crashkernel, it may cause OOM in the first kernel
> 
> Appreciate for your review!

My pleasure! Thanks for addressing my feedback!

coiby (changes requested):

kdumpctl
```diff
@@ -1129,6 +1129,53 @@ prepare_luks()
...
+
+suggest_crashkernel_reset()
+{
+	local _crashkernel _recommend_value _actual_value _bounce_buff_size
+
+	_crashkernel=$(sed -n 's/.*crashkernel=\([^ ]*\).*/\1/p' /proc/cmdline)
+	[[ $(kdump_get_conf_val auto_reset_crashkernel) == no ]] && return
+
+	if ! echo "$_crashkernel" | grep -q -E "(fadump|fadump=0)"; then
+		_recommend_value=$(kdump_get_arch_recommend_size)
```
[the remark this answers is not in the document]
pfliu: Ah, it is based on the fact that the crashkernel value for fadump is not dynamic. We rely on the value provided by IBM. I have added a note in the code.

Also I have fixed the code as your suggestion.

Thanks for your time.

PR #93, opened 2025-06-29
kdump.conf add dracut_kdump_service_timeout

jinkkkang:
The timeout time of dracut kdump service is currently 300 seconds. When i test on a physical machine with 2T memory, this time is not enough. Therefore, this  dracut_kdump_service_timeout parameter is added to support custom configuration timeout

## Summary by Sourcery

Add support for configuring the dracut kdump service start timeout via a new kdump.conf parameter and apply it to systemd’s DefaultTimeoutStartSec.

New Features:
- Introduce dracut_kdump_service_timeout in kdump.conf to allow custom kdump service start timeouts.

Enhancements:
- Propagate the custom timeout value (or fallback to default 300 s) into systemd’s kdump service configuration.

Documentation:
- Document the dracut_kdump_service_timeout parameter and its default value in the kdump.conf man page.

### DIFF
--- a/dracut/99kdumpbase/module-setup.sh
+++ b/dracut/99kdumpbase/module-setup.sh
@@ -1051,10 +1051,11 @@ kdump_install_random_seed() {
 kdump_install_systemd_conf() {
     # Kdump turns out to require longer default systemd mount timeout
     # than 1st kernel(45s by default), we use default 300s for kdump.
+    kdump_timeout=$(get_kdump_service_timeout)
     mkdir -p "${initdir}/etc/systemd/system.conf.d"
     cat > "${initdir}/etc/systemd/system.conf.d/99-kdump.conf" << EOF
 [Manager]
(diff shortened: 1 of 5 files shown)


coiby:
Hi @jinkkkang,

Thanks for reporting your problem and also sending an PR to fix it! 

Quoting commit 2777a93a ("mkdumprd: use 300s as the default systemd unit timeout for kdump mount"), 
> "DefaultTimeoutStartSec=X" can be overridden by individual mount option "x-systemd.device-timeout=X", users can specify their own values as needed.

So if `x-systemd.device-timeout=X` works for your case, I'd prefer to not add a new kdump.conf option and simply improve kdump.conf.5 to help users be aware of this solution.

coiby:
Hi @jinkkkang,

Can you provide the console log if it's available to help me understand better about this problem? Thanks!

PR #98, opened 2025-07-23
Support dumping to NVMe/TCP configured using NVMe Boot Firmware Table

coiby:
The dracut nvmf module can take care of all things. It can parse ACPI
NVMe Boot Firmware Table (NBFT) tables, generate NetworkManager profiles
and discover and connect all subsystems.

Currently, the dracut kdump module will try to bring up the same network
connections as in 1st kernel. But a different set of NVMe connections
and active network interfaces will be used for the case of multipathing.
So the dracut kdump module should let dracut nvmf module do everything.

Note connecting everything and having network redundancy may require extra
memory and the default crashkernel may not work. We'll document this
issue and ask users to increase the crashkernel.

### DIFF
--- a/dracut/99kdumpbase/module-setup.sh
+++ b/dracut/99kdumpbase/module-setup.sh
@@ -360,11 +360,6 @@ kdump_install_nmconnections() {
             exit 1
         fi
     done <<< "$(nmcli -t -f device,filename connection show --active)"
-
-    # Stop dracut 35network-manger to calling nm-initrd-generator.
-    # Note this line of code can be removed after NetworkManager >= 1.35.2
-    # gets released.
-    echo > "${initdir}/usr/libexec/nm-initrd-generator"
 }
 
 kdump_install_nm_netif_allowlist() {
@@ -658,7 +653,7 @@ kdump_install_net() {
         kdump_install_nmconnections
         apply_nm_initrd_generator_timeouts
         kdump_setup_znet "$_netifs"
-        kdump_install_nm_netif_allowlist "$_netifs"
+        [[ $is_nvmf ]] || kdump_install_nm_netif_allowlist "$_netifs"
         kdump_install_nic_driver "$_netifs"
         kdump_install_resolv_conf
     fi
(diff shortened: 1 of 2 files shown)


dracut/99kdumpbase/module-setup.sh
```diff
@@ -915,6 +914,33 @@ kdump_check_iscsi_targets() {
...
+    for _d in device/nvme*; do
+        [ -L "$_d" ] || continue
+        if readlink "$_d" | grep -q nvme-fabrics; then
+            read -r _trtype < "$_d"/transport
+            [[ $_trtype == "fc" || $_trtype == "tcp" || $_trtype == "rdma" ]] && return 0
+        fi
+    done
+    return 1
+}
+
```
tbzatek, line 939: Well, having an `nvme-fabrics`-class block device present in the system doesn't mean it was connected during boot. There are multiple points where NVMeoF devices can get connected. There's currently no block device tagging in place (might come in the future).

Perhaps checking for `/sys/firmware/acpi/tables/NBFT*` tables presence might be more precise for Boot from NVMe/TCP?

tbzatek: Right, the configured kdump target may not be connected at boot, so this code is still needed for such a scenario and at the same time it's difficult to distinguish the devices from NBFT-connected ones (unless you parse the NBFT table yourselves, but even then such detection is not 100% reliable due to the possibility of NVMe discovery initiated from NBFT). And you're right, the `nvmf` dracut module serves three purposes: NVMe/FC boot, NVMe/TCP boot (NBFT) and the `/etc/nvme/discover.conf` legacy way.

Thinking of it, theoretically generating `/etc/nvme/discover.conf` file from an actual system state might be another way. The good thing about `nvme-cli connect` command is that it first tries to match existing connections and will skip the connection attempt in such a case. So connecting NBFT first and then the `/etc/nvme/discover.conf` entries would still play nice.

Either way, the `/etc/nvme/discover.conf` way of connecting devices has not been properly tested and is not a priority at the moment I would say.

tbzatek (approved):
Anyway, I don't want to block the progress, this looks good enough for the moment, let's deal with further issues later.

dracut/99kdumpbase/module-setup.sh
coiby: I see. Let's support the /etc/nvme/discover.conf way of connecting in the future then.

coiby:
Thanks for approving the PR! I will merge this this PR without the 2nd commit [Disable dracut nvmf module if the dumping target is not NVMe over Fab… · rhkdump/kdump-utils@041ed35](https://github.com/rhkdump/kdump-utils/commit/041ed3561429e3b5723edb437a1e06328a225c17) which is useless as it still uses for_each_host_dev_and_slaves_all.

PR #101, opened 2025-07-31
Try to bring up the most recently used NetworkManager connection profiles

coiby:
Try to bring up the most recently used NetworkManager connection profiles

Resolves: https://issues.redhat.com/browse/KEX-344

Currently, the nfs_ovs plan fails because nm-wait-online-initrd.service
waits for the bonding network to be created. But the bonding network
can't be activated because the eth0 NM connection is brought up first
which stops the slave connnection profile to be activated,

/var/lib/NetworkManager/timestamps keeps the timestamps of NM
connections when they are lastly used. When multiple NM connections have
the same connection.autoconnect-priority, the latest used NM connection
will be activated.

Note this PR also enable tests for rawhide and aarch64.

[1] https://networkmanager.dev/docs/api/latest/settings-connection.html

Fixes: 3f4a9eb4 ("Bring up the network before Open vSwitch bridge")


licliu (approved):
LGTM!

PR #102, opened 2025-08-11
sysconfig: disable kfence in kdump kernel

baoquan-he:
In commit ddb0bab1f7e1, I mistakenly added the 'kfence.sample_interval=0' into KDUMP_COMMANDLINE_APPEND which is taken as default parameters added into kdump kernel. The KDUMP_COMMANDLINE_APPEND value is for unsupported architecuture. We should append wanted kernel parameter to each supported architecture in "update_param KDUMP_COMMANDLINE_APPEND" place.

Fix it now.


licliu (approved):
LGTM.

BTW, this issue was not found in review, maybe we can do some improvements for default case in the future.

PR #107, opened 2025-09-17
Add kdumpctl setup-crypttab subcommand

coiby:
Resolves: https://issues.redhat.com/browse/RHEL-29037

Relates: https://issues.redhat.com/browse/RHEL-29039

This subcommand is to add the 'link-volume-key' option to /etc/crypttab
so the volume keys can be passed to the crash kernel to unlock
LUKS-encrypted device automatically.

This API will be also be called by kdump-anaconda-addon to set up
/etc/crypttab on installation.

### DIFF
--- a/dracut/99kdumpbase/kexec-crypt-setup.sh
+++ b/dracut/99kdumpbase/kexec-crypt-setup.sh
@@ -1,3 +1,21 @@
 #!/bin/sh
 #
-echo true > /sys/kernel/config/crash_dm_crypt_keys/restore
+LUKS_CONFIGFS_RESTORE=/sys/kernel/config/crash_dm_crypt_keys/restore
+RESTORED=1
+MAX_WAIT_TIME=10
+wait_time=0
+
+while [ $wait_time -lt $MAX_WAIT_TIME ]; do
+    [ -e $LUKS_CONFIGFS_RESTORE ] && break
+    sleep 1
+    wait_time=$((wait_time + 1))
+done
+
+if [ $wait_time -ge $MAX_WAIT_TIME ]; then
+    echo "$LUKS_CONFIGFS_RESTORE isn't ready after ${MAX_WAIT_TIME}s, something wrong!"
+    exit 1
+fi
+
+if ! grep -q "$RESTORED" "$LUKS_CONFIGFS_RESTORE"; then
+    echo $RESTORED > $LUKS_CONFIGFS_RESTORE
+fi
--- a/dracut/99kdumpbase/module-setup.sh
+++ b/dracut/99kdumpbase/module-setup.sh
@@ -1095,7 +1095,7 @@ EOF
 }
 
 kdump_check_crypt_targets() {
-    local _luks_dev _devuuid _key_desc
+    local _devuuid _key_desc
     declare -a _luks_devs
 
     mapfile -t _luks_devs < <(get_all_kdump_crypt_dev)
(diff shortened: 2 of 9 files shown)


daveyoung:
> > > Hi @daveyoung,
> > > Does the recent doc change look good to you?
> > 
> > 
> > Hi @coiby Looks good. But I have another question about overwriting the cryptab file, it maybe a silly question as I did not use it at all and I know nothing about LUKs. Is it possible to put the kdump related configuration options to other config file dir eg. /etc/cryptsetup-keys.d/ or /run/cryptsetup-keys.d/? So reduce the risk of causing 1st kernel boot failure if some bug exists in our code.
> 
> Unfortunately I don't think crypttab has provided any drop-in mechanism. A close one is to add luks.options kernel cmdline parameter but it will be become quite lengthy e.g. `luks.options=ed8de5d4-1975-4357-85eb-077c02ef4aaf=link-volume-key=@u::%logon:kdump-cryptsetup-vk:ed8de5d4-1975-4357-85eb-077c02ef4aaf`. If there are multiple LUKS devices used for kdump, then the kernel cmdline will become difficult to read.
> 
> So I choose to modify crypttab and use a thorough set of unit tests to ensure the code robustness.

Ok, I'm not rejecting the pr, just explore if there are something missed.  saying about kernel cmdline,  maybe check https://www.kernel.org/doc/Documentation/admin-guide/bootconfig.rst,  is it suitable for your use case?

coiby:
> > > > > Hi @daveyoung,
> > > > > Does the recent doc change look good to you?
> > > > 
> > > > 
> > > > Hi @coiby Looks good. But I have another question about overwriting the cryptab file, it maybe a silly question as I did not use it at all and I know nothing about LUKs. Is it possible to put the kdump related configuration options to other config file dir eg. /etc/cryptsetup-keys.d/ or /run/cryptsetup-keys.d/? So reduce the risk of causing 1st kernel boot failure if some bug exists in our code.
> > > 
> > > 
> > > Unfortunately I don't think crypttab has provided any drop-in mechanism. A close one is to add luks.options kernel cmdline parameter but it will be become quite lengthy e.g. `luks.options=ed8de5d4-1975-4357-85eb-077c02ef4aaf=link-volume-key=@u::%logon:kdump-cryptsetup-vk:ed8de5d4-1975-4357-85eb-077c02ef4aaf`. If there are multiple LUKS devices used for kdump, then the kernel cmdline will become difficult to read.
> > > So I choose to modify crypttab and use a thorough set of unit tests to ensure the code robustness.
> > 
> > 
> > Ok, I'm not rejecting the pr, just explore if there are something missed. saying about kernel cmdline, maybe check https://www.kernel.org/doc/Documentation/admin-guide/bootconfig.rst, is it suitable for your use case?
> 
> Thanks for the suggestion! Wow, bootconfig is to totally new to me. I just had a quick look at bootconfig and think at least it will be free of the issue of length kernel cmdline in /proc/cmdline. But one limitation is [systemd-cryptsetup-generator](https://github.com/systemd/systemd/blob/26cde6f0cefab1acc817c01e8dbe935c40db1872/src/cryptsetup/cryptsetup-generator.c#L167) only read /proc/cmdline and not /proc/bootconfig. I will need more time to investigate this approach is doable and better. I need to find out 1) if systemd community will accept the solution of also reading LUKS related configurations from bootconfig 2) if anaconda accepts modifying built initramfs on OS installation and 3) find a solution to build and package linux/tools/bootconfig/bootconfig.

On second thought, bootconfig is much more complex than crypttab (two mandatory fields + two optional fields separated by space). Shall we implement the C code in systemd to parse /proc/bootconfig, it will be much more work. So I think it's not worth adopting the bootconfig approach.

Issue #113, opened 2025-09-24
Unable to find the crash kernel when UKI is used


litian1992:
The CVM images created by Image Builder have to have the UKI named like this in ESP.
`ffffffffffffffffffffffffffffffff-6.12.0-125.el10.x86_64.efi`
Although it's supposed to be machine ID at its front. IB developer claims machine ID is unknown before the creation of VM thus unable to name it properly.
This in turn causes kdump unable to find kernel by referencing `/etc/machine-id`
https://github.com/rhkdump/kdump-utils/blob/main/kdump-lib.sh#L522
The EFI var `/sys/firmware/efi/efivars/LoaderImageIdentifier-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f` can be used instead.

PR #114, opened 2025-09-24
kdump-lib: use EFI var to find UKI

litian1992:
Resolves: #113


coiby:
Hi @litian1992,

I don't see this patch in RHEL. Does it need to be backported to RHEL?

PR #115, opened 2025-10-09
refactor(gen-kdump-sysconfig): Use bash heredoc and associative arrays

coiby:
The script gen-kdump-sysconfig.sh is refactored to use bash heredoc instead of regex to generate kdump sysconfig file. Bash associative arrays are also used to manage architecture-specific values. This change improves readability and maintainability by centralizing the configuration for each architecture. For example, we could have avoided the mistake made by commit ddb0bab1 ("sysconfig: disable kfence in kdump kernel").

The default values are now stored under the 'default' key within the associative arrays, simplifying the logic for handling unknown or unspecified architectures.

Assisted-by: Google Gemini

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,5 +18,5 @@ binary_next_line   = false
 space_redirects    = true
 
 # Some scripts will only run with bash
-[{mkfadumprd,mkdumprd,kdumpctl,kdump-lib.sh}]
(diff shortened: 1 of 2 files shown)


coiby:
Hi @prudo1 

Thanks for reviewing the PR! I've applied Gemini's improvements and also your suggestion

prudo1 (approved):
Hi Coiby,
looks good to me!
Thanks
Philipp

Issue #108, opened 2025-09-18
logger function like dinfo turn off xtrace unless environment variable debug set


coiby:
logger functions like dinfo will disable xtrace unexpectedly,
```bash
ddebug()
{
	set +x
	dlog 4 "$@"
	if [ -n "$debug" ]; then
		set -x
	fi
}

dinfo()
{
	set +x
	dlog 3 "$@"
	if [ -n "$debug" ]; then
		set -x
	fi
}

```
As a result, `bash -x kdumpctl` or `shellspec -x` may show incomplete xtrace. I'm not sure if we should re-enable xtrace automatically if we detect `x` in `"$-"`

prudo1:
Hi @coiby,
I believe that is something we should fix. I noticed that the same problem occurs when adding `rd.debug` to the kernel command line, which enables xtrace for the services in the initrd. This can be very useful to debug problems within the kdump initrd.

coiby:
Hi @prudo1 ,

Thanks for sharing you thoughts! I've created #126

PR #117, opened 2025-10-16
Print CORE_COLLECTOR as debug info

sourabhjains:
Printing CORE_COLLECTOR helps developers identify which core collector, along with its respective options, was used to capture the dump.

It is also helpful when writing kdump and kdump-utils tests using external test frameworks like op-tests.

### DIFF
--- a/dracut/99kdumpbase/kdump.sh
+++ b/dracut/99kdumpbase/kdump.sh
@@ -182,6 +182,7 @@ dump_fs() {
     KDUMP_LOG_DEST=$_dump_fs_path/
     KDUMP_LOG_OP="mv '$KDUMP_LOG_FILE' '$KDUMP_LOG_DEST/'"
 
+    ddebug "Core Collector: $CORE_COLLECTOR"
     $CORE_COLLECTOR /proc/vmcore "$_dump_fs_path/vmcore-incomplete" > /dev/console 2>&1
     _dump_exitcode=$?
     if [ $_dump_exitcode -eq 0 ]; then
@@ -392,6 +393,7 @@ dump_raw() {
     fi
 
     dinfo "saving vmcore"
+    ddebug "Core Collector: $CORE_COLLECTOR"
     $CORE_COLLECTOR /proc/vmcore | dd of="$1" bs=$DD_BLKSIZE >> /tmp/dd_progress_file 2>&1 || return 1
     sync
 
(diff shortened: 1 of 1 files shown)


sourabhjains:
> Hi Sourabh,
> 
> I don't understand the problem you are trying to solve with this PR. Can you please explain in what situation printing of the `CORE_COLLECTOR` helps and

First, while debugging kdump, one can see which CORE_COLLECTOR along with its options was used to capture the dump.

Second, to automate the kdump and kdump-utils tests, if CORE_COLLECTOR is printed on the console, one can parse it to verify that the configured CORE_COLLECTOR option was actually used during dump capture.

> why it needs special handling compared to the other config options.

I just want to print the CORE_COLLECTOR during dump collection to see which options were used by the CORE_COLLECTOR. For example, what was the value of --num-threads= for the makedumpfile CORE_COLLECTOR.

prudo1:
Hi Sourabh,

> > Hi Coiby, Hi Sourabh,
> > sorry for the late reply.
> > I still don't see the benefit in adding those debug messages. For me it looks like a workaround without addressing the underlying problem. I'd prefer see a cleanup so that the `CORE_COLLECTOR` isn't changed for every possible dump target instead of just printing the final result. Especially as the `--num-threads` implementation is broken for raw dumps in the moment. It combines `--num-threads` with `-F` which IIUC shouldn't be possible.
> 
> I understand there might be some confusion around CORE_COLLECTOR, as we don’t yet have a common function to manage it.
> 
> That said, this patch only adds a print for CORE_COLLECTOR, and that too only when debug mode is enabled. The kdump service already prints several details, including its log level, but it currently misses one key piece of information - the CORE_COLLECTOR. I believe adding this would help both while debugging and during test automation.

you are not only adding a print. You are also adding a dependency of an external test on that print, which isn't documented anywhere. So this print adds a big maintenance overhead for us in the long term. That's why I want to have a more generic solution that helps with not just this single problem. 

All in all the best way forward is to fix the xtrace issue @coiby  is tracking in #108. When this is fixed you can add `rd.debug` to the kernel command line and get the full xtrace output, i.e. every line of  bash code executed in the initrd. Fixing this will help with debugging a lot of issues in the kdump initrd, including yours.

Thanks
Philipp

sourabhjains:
I will try the changes introduce in #108 and close this pull request.

coiby:
Hi @sourabhjains,

> I will try the changes introduce in #108 and close this pull request.

Can you give [kdump-utils-1.0.60-1.fc45](https://koji.fedoraproject.org/koji/buildinfo?buildID=2947574) a try?

PR #121, opened 2025-10-30
Add persisent device if FIPS is enabled

hvj7:
mkdumprd has a code to add a disk to kdump initramfs, in case FIPS is enabled and /boot is on a separate partition. This code used to work, since dracut was not force checking that added disk is in fact available. Since dracut commit c79fc8f dracut in fact checks for added device, and since disk name could have been changed between live system and kdump initramfs, kdump can fail. To resolve this problem add disk by UUID, not by disk name.

### DIFF
--- a/mkdumprd
+++ b/mkdumprd
@@ -433,8 +433,9 @@ if ! is_fadump_capable; then
 	# to /sysroot beforehand.
 	if [[ $(cat /proc/sys/crypto/fips_enabled 2> /dev/null) == 1 ]]; then
 		_boot_source=$(findmnt -n -o SOURCE --target /boot)
+		_disk_persistent=$(get_persistent_dev "$_boot_source")
 		if mountpoint -q /boot; then
-			dracut_args+=(--add-device "$_boot_source")
+			dracut_args+=(--add-device "$_disk_persistent")
 		else
 			add_mount "$_boot_source"
 		fi


coiby:
Hi @hvj7,

Thanks for creating this PR!

If I understand [the dracut commit c79fc8f](https://github.com/dracut-ng/dracut-ng/commit/c79fc8f) correctly, it will make systemd wait infinitely for a device not specified by persistent names e.g. /dev/sda1. So I don't think it will cause a kdump failure. On the contrary, this dracut commit can address the failure when a block device needs more than 1min 30s to be activated. If I miss anything, please let me know. Thanks!

coiby:
Hi @hvj7 and @aburmash,

Finally I understand what problem [the dracut commit c79fc8f](https://github.com/dracut-ng/dracut-ng/commit/c79fc8f) tries to resolve. Before that commit, with `--add-device /dev/mapper/vg-lvol0`, dev-vg-lvol0.device can still time out despite that /dev/mapper/vg-lvol0  is already a persistent name. Because dracut would wait forever for a different persist name e.g. /dev/disk/by-uuid/69f27553-5f60-41e8-94a7-51e6b8da79fc instead. I may send a PR to update the dracut's doc.

Meanwhile, I think there is a necessity to  make a change in kdump-utils/mkdrump. Can you update the PR to re-use `get_persistent_dev` from dracut-functions.sh which will be sourced by mkdrumpd to get persistent device name instead?

coiby:
> > > Please let me know if this commit message is fine @coiby
> > 
> > 
> > Thanks for updating the commit message! Once the subject line gets updated as well, we are good to go:)
> 
> Ah snap forgot about the subject line. Let me know if it's okay now.

Thanks! I notice the static-analysis test fails because a `fi` gets removed by mistake. Sorry the test failed to run automatically seemingly caused a permission issue.

PR #122, opened 2025-11-04
kdump.sh: Strip the surrounding quotes from configuration values

licliu:
commit 1a733872(Strip surrounding quotes from configuration values) omitted modifications to kdump.sh.

Adds a POSIX-compliant `case` statement to remove the surrounding quotes.

### DIFF
--- a/spec/kdump-lib-initramfs_spec.sh
+++ b/spec/kdump-lib-initramfs_spec.sh
@@ -12,6 +12,11 @@ Describe 'kdump-lib-initramfs'
 			#|dracut_args --omit-drivers "cfg80211 snd" --add-drivers "ext2 ext3"
 			#|sshkey /root/.ssh/kdump_id_rsa
 			#|ssh user@my.server.com
+			#|    ext2	  	/dev/sda1		# trailing comment
+			#|    ext2		/dev/sdb1	   	# trailing comment
+			#|virtiofs /dev/sdc1
+			#|# Note: this and the next line are comments
+			#|#virtiofs /dev/sdd1
 			#|core_collector "makedumpfile -l --message-level 7 -d 31"
 		}
 		kdump_config >$KDUMP_CONFIG_FILE
@@ -23,14 +28,20 @@ Describe 'kdump-lib-initramfs'
 			#  - Given two parameters, retrive one parameter that has value specified
 			#  - Given two parameters (in reverse order), retrive one parameter that has value specified
 			#  - values are enclosed in quotes
+			#  - when the same option is set multiple times, return the last entry
+			#  - comments are treated properly
+			#  - when a option is not set, return 0 and an empty string
 			Parameters
 				"#1" nfs my.server.com:/export/tmp
 				"#2" ssh user@my.server.com
 				"#3" failure_action shell
 				"#4" dracut_args '--omit-drivers "cfg80211 snd" --add-drivers "ext2 ext3"'
-				"#5" 'ssh\|aaa' user@my.server.com
-				"#6" 'aaa\|ssh' user@my.server.com
+				"#5" 'ssh|aaa' user@my.server.com
+				"#6" 'aaa|ssh' user@my.server.com
 				"#7" core_collector "makedumpfile -l --message-level 7 -d 31"
+				"#8" ext[234] /dev/sdb1
+				"#9" 'ext[234]|xfs|btrfs|minix|raw|virtiofs' /dev/sdc1
+				"#10" foo ""
 			End
 
 			It 'should handle all cases correctly'


licliu:
Thanks @prudo1, I think you've done most of the work, I just added some minor changes and commit messages.

dracut/99kdumpbase/module-setup.sh
```diff
@@ -724,7 +724,7 @@ kdump_install_conf() {
     kdump_read_conf > "${initdir}/tmp/$$-kdump.conf"
 
     while read -r _opt _val; do
-        [[ $_val == \"*\" ]] && _val=${_val:1:-1}
+        # remove inline comments after the end of a directive.
```
prudo1: Hi @licliu,

I agree with reverting the commit. But could you please add a short reason to the commit message so that it's easier to understand in the future why the commit was reverted

Thanks
Philipp

kdump-lib-initramfs.sh
```diff
@@ -19,19 +19,49 @@ LUKS_KEY_PRFIX="kdump-cryptsetup:vk-"
...
-	# Following steps are applied in order: strip trailing comment, strip trailing space,
-	# strip heading space, match non-empty line, remove duplicated spaces between conf name and value
-	[ -f "$KDUMP_CONFIG_FILE" ] && sed -n -e "s/#.*//;s/\s*$//;s/^\s*//;s/\(\S\+\)\s*\(.*\)/\1 \2/p" $KDUMP_CONFIG_FILE
+	kdump_get_conf_val ""
 }
 
 # Retrieves config value defined in kdump.conf
-# $1: config name, sed regexp compatible
+# $1: config name, if empty print full config
 kdump_get_conf_val()
```
prudo1: Tiny nit, there are at least two users of `kdump_get_conf_val` that use a regex to check multiple configs at once. In particular, `is_user_configured_dump_target` and `get_block_dump_target`. Those users need to be updated as well to work properly. Please also double check that I didn't miss any other user.

kdump-lib-initramfs.sh
```diff
@@ -19,19 +19,49 @@ LUKS_KEY_PRFIX="kdump-cryptsetup:vk-"
...
+		if [ -z "$_to_find" ]; then
+			echo "$_opt $_val"
+		elif echo "$_opt" | grep -q -E "^($_to_find)$"; then
+			# make sure to only return the last match to mirror the
+			# old behavior
+			_found="$_val"
+		fi
+	done < "$KDUMP_CONFIG_FILE"
+	[ -n "$_found" ] && echo "$_found"
 }
```
prudo1: Hi @licliu,
I found an other nit. In case an option is queried that is not set the old behavior was to return 0 and an empty string. But the new code will return 1, the result of `[ -n $_found ]`, and an empty string. So we need to explicitly return 0 at the end of the function.

I found this by adding a test to the test case. So it's probably best to keep it.

All in all the changes I made are the following

```
diff --git a/kdump-lib-initramfs.sh b/kdump-lib-initramfs.sh
index 3639be6..6f5d6db 100755
--- a/kdump-lib-initramfs.sh
+++ b/kdump-lib-initramfs.sh
@@ -62,6 +62,9 @@ kdump_get_conf_val()
                fi
        done < "$KDUMP_CONFIG_FILE"
        [ -n "$_found" ] && echo "$_found"
+
+       # make sure we return 0 even when a option isn't set
+       return 0
 }
 
 is_mounted()
diff --git a/spec/kdump-lib-initramfs_spec.sh b/spec/kdump-lib-initramfs_spec.sh
index 59bfb88..18dc060 100644
--- a/spec/kdump-lib-initramfs_spec.sh
+++ b/spec/kdump-lib-initramfs_spec.sh
@@ -28,6 +28,9 @@ Describe 'kdump-lib-initramfs'
                        #  - Given two parameters, retrive one parameter that has value specified
                        #  - Given two parameters (in reverse order), retrive one parameter that has value specified
                        #  - values are enclosed in quotes
+                       #  - when the same option is set multiple times, return the last entry
+                       #  - comments are treated properly
+                       #  - when a option is not set, return 0 and an empty string
                        Parameters
                                "#1" nfs my.server.com:/export/tmp
                                "#2" ssh user@my.server.com
@@ -36,8 +39,9 @@ Describe 'kdump-lib-initramfs'
                                "#5" 'ssh|aaa' user@my.server.com
                                "#6" 'aaa|ssh' user@my.server.com
                                "#7" core_collector "makedumpfile -l --message-level 7 -d 31"
-                               "#8" ext2 /dev/sdb1
+                               "#8" ext[234] /dev/sdb1
                                "#9" 'ext[234]|xfs|btrfs|minix|raw|virtiofs' /dev/sdc1
+                               "#10" foo ""
                        End
 
                        It 'should handle all cases correctly'
```

With that fixed I believe the PR is good to merge.

Thanks
Philipp

kdump-lib-initramfs.sh
```diff
@@ -19,19 +19,49 @@ LUKS_KEY_PRFIX="kdump-cryptsetup:vk-"
...
-	# For lines matching "^\s*$1\s+", remove matched part (config name including space),
-	# remove tailing comment, space, then store in hold space. Print out the hold buffer on last line.
-	[ -f "$KDUMP_CONFIG_FILE" ] &&
-		sed -n -e "/^\s*\($1\)\s\+/{s/^\s*\($1\)\s\+//;s/#.*//;s/\s*$//;h};\${x;p}" $KDUMP_CONFIG_FILE
+	_to_find="$1"
+	_found=""
+
+	[ -f "$KDUMP_CONFIG_FILE" ] || return
+	while read -r _line; do
+		_line="$(echo "$_line" | tr -s "[:blank:]" " ")"
```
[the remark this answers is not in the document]
prudo1: Yes, I don't believe it's an important issue for us. But still it's a limitation we need to keep in mind. Just in case some user complains.

PR #123, opened 2025-11-05
Some fixes on kdump LUKS support

coiby:
This PR fixes three situations where kdump can fail by
 1. adding "KeyringMode=shared" to the kdump.service unit file so it can access the LUKS volume keys
 2. fixing the SELinux label of crypttab file
 3. allowing users to use "sudo kdumpctl"

### DIFF
--- a/kdumpctl
+++ b/kdumpctl
@@ -737,7 +737,8 @@ function load_kdump_kernel_key()
 	if ! [[ -f /proc/device-tree/ibm,secure-boot ]]; then
 		return
 	fi
-
+	# %.ima keyring is writable to the user, no need to use
+	# "sudo -i keyctl"
 	keyctl padd asymmetric "" %:.ima < "/usr/share/doc/kernel-keys/$KDUMP_KERNELVER/kernel-signing-ppc.cer"
 }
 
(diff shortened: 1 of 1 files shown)


coiby:
> Hi @coiby, I found two places in the code that use `keyctl` instead of `KEYCTL_CMD`. In my tests, they didn't make bad result. But do you think they should be replaced with `KEYCTL_CMD` to maintain consistency?

Hi @licliu, thanks for raising the concern! 

> 
> ```
> kdumpctl:741:   keyctl padd asymmetric "" %:.ima < "/usr/share/doc/kernel-keys/$KDUMP_KERNELVER/kernel-signing-ppc.cer"
> kdumpctl:1107:  done < <(keyctl list @u 2> /dev/null || true)
> ```

There is no need to use "sudo -i" in these two places so there is no need to use `KEYCTL_CMD`. I've added some code comments and hopefully it can provide more clarity. If there is still something confusing, please let me know.

> 
> For the other parts, this pr looks good to me.

Thanks for reviewing the PR!

licliu (approved):
Thank you for your explanation!

PR #128, opened 2025-11-28
Add systemd service for servicelog_notify

licliu:
This introduces a new systemd service for servicelog_notify to ensure proper setup during the actual boot process (e.g., for bootc environments).

This commit only provides the unit file and has no immediate operational impact. Downstream packaging must configure the appropriate systemd preset and post-install script to enable it.

Unlike the previous method, which only executed during kdump-utils installation or upgrades, via a systemd service will run servicelog_notify on every system boot.

### DIFF
--- a/kdump-migrate-notify.service
+++ b/kdump-migrate-notify.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Register kdump migration notification
+After=kdump.service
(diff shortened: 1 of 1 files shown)


licliu:
Perhaps we could consider enabling it only on ostree-based systems by checking `ConditionPathExists=/run/ostree-booted`?

PR #129, opened 2025-12-01
kexec-kdump-howto.txt: update paragraphs related to disable_cpu_apicid

baoquan-he:
Long before, to support multiple CPUs on x86_64, 'disable_cpu_apicid='
was introduced. It's to avoid the case when nr_cpus=xx is added, while
crashed cpu is not BSP cpu, then the crashed CPU will send INIT to BSP 
cpu in kdump kernel. While the BSP cpu being reinitialized when
receiving INIT in the 2nd time will cause kdump kernel collapsing.

Now, in kernel commit 5c5682b9f87a ("x86/cpu: Detect real BSP on crash
kernels"), the requirement of disable_cpu_apicid has been taken off by
detecting real BSP on crashed kernel and not sending INIT to it. 
Now testing passed on x86_64 system w/ or w/o disable_cpu_apicid and the 
kernel message can be seen as below.

===
CPU topo: Boot CPU APIC ID not the first enumerated APIC ID: 1e != 0
CPU topo: Crash kernel detected. Disabling real BSP to prevent machine INIT
===

However, sometime older kernels which don't contain commit 5c5682b9f87a
are still loaded in newer OS. To keep good back compatibility, still
adding disable_cpu_apicid to capture kernel. It's not harmful on newer
kernels containing commit 5c5682b9f87a.

### DIFF
--- a/kexec-kdump-howto.txt
+++ b/kexec-kdump-howto.txt
@@ -935,10 +935,14 @@ For example:
 
 Notes on how to use multiple cpus on a capture kernel on x86 system:
 
-Make sure that you are using a kernel that supports disable_cpu_apicid
-kernel option as a capture kernel, which is needed to avoid x86 specific
-hardware issue (*). The disable_cpu_apicid kernel option is automatically
-appended by kdumpctl script and is ignored if the kernel doesn't support it.
+The disable_cpu_apicid kernel option is automatically appended by kdumpctl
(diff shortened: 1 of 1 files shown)


coiby:
Hi @baoquan-he,

The PR LGTM! And I think the two minor suggestions from Gemini can help improve readability, what do you think?

PR #133, opened 2025-12-31
kdump-lib: Retrieve xDUMP_COMMANDLINE_APPEND from sysconfig

pfliu:
The current references to KDUMP_COMMANDLINE_APPEND or FADUMP_COMMANDLINE_APPEND return nothing because these variables are dynamically generated by gen-sysconfig-kdump.sh.

To retrieve these variables, introduce a function
get_dump_commandline(). It is the caller's responsibility to ensure '/etc/sysconfig/kdump' exists; otherwise, get_dump_commandline() will output an error message.

### DIFF
--- a/kdump-lib.sh
+++ b/kdump-lib.sh
@@ -16,6 +16,8 @@ FADUMP_REGISTER_SYS_NODE="/sys/kernel/fadump/registered"
 FADUMP_APPEND_ARGS_SYS_NODE="/sys/kernel/fadump/bootargs_append"
 # shellcheck disable=SC2034
 FENCE_KDUMP_CONFIG_FILE="/etc/sysconfig/fence_kdump"
+# shellcheck disable=SC2034
+KDUMP_SYSCONFIG_FILE="/etc/sysconfig/kdump"
 
 is_fadump_capable()
 {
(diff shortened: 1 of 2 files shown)


coiby:
Hi @pfliu,

This change is only for find_nr_cpus. But the only caller of find_nr_cpus is kdump_get_arch_recommend_crashkernel from kdumpctl. Since kdumpctl alreadys kdumpctl sources /etc/sysconfig/kdump, it seems there is no need to parse  /etc/sysconfig/kdump. Or do I miss anything?

pfliu:
> > > Hi @pfliu,
> > > This change is only for find_nr_cpus. But the only caller of find_nr_cpus is kdump_get_arch_recommend_crashkernel from kdumpctl. Since kdumpctl alreadys kdumpctl sources /etc/sysconfig/kdump, it seems there is no need to parse /etc/sysconfig/kdump. Or do I miss anything?
> > 
> > 
> > I ignored the source "/etc/sysconfig/kdump" in kdumpctl. I think that moving this source into kdump-lib.sh can resolve the issue.
> 
> By "the issue", do you mean "The current references to KDUMP_COMMANDLINE_APPEND or FADUMP_COMMANDLINE_APPEND return nothing"? Since kdumpctl will source /etc/sysconfig/kdump, this issue won't happen, right?

Yes, you are right. I made a mistake when reading the code. I will close the PR.

Thanks,

Pingfan

PR #142, opened 2026-01-22
kdump-lib: check kdump service state when validating fadump load

sourabhjains:
Currently, kdumpctl status reports kdump as operational even when the kdump service is in a failed state for fadump.

root> kdumpctl status
kdump: Dump mode is fadump
kdump: Kdump is operational

root> systemctl status kdump
× kdump.service - Crash recovery kernel arming
     Loaded: loaded (/usr/lib/systemd/system/kdump.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Tue 2026-01-20 02:52:59 CST; 12s ago
<snip...>
cp: error writing '/boot/initramfs-6.12.0-180.el10.ppc64le.img.tmp': No space left on device dracut[F]: Creation of /boot/initramfs-6.12.0-180.el10.ppc64le.img.tmp failed <snip...>

This happens because the kernel registers fadump during early boot [1]. Even if the kdump service later fails to start due to a failure while updating the initrd with the fadump component, /sys/kernel/fadump/registered still remains set to 1. As a result, kdump is reported as operational, even though it is not.

Fix this by checking the kdump service state when validating the fadump load.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=607451ce0aa9bdff590db4d087173edba6d7a29d

Reported-by: Shirisha G <shirisha@linux.ibm.com>

### DIFF
--- a/kdumpctl
+++ b/kdumpctl
@@ -2209,6 +2209,7 @@ main()
 {
 	# Determine if the dump mode is kdump or fadump
 	determine_dump_mode
+	local rc
 
 	case "$1" in
 	start)
@@ -2228,7 +2229,13 @@ main()
 	status)
 		EXIT_CODE=0
 		is_kernel_loaded "$DEFAULT_DUMP_MODE"
-		case "$?" in
+		rc=$?
+
+		if [ "$rc" -eq 0 ] && [ "$DEFAULT_DUMP_MODE" = "fadump" ]; then
+			systemctl is-active --quiet kdump.service || rc=1
+		fi
+
+		case "$rc" in
 		0)
 			dinfo "Kdump is operational"
 			EXIT_CODE=0


sourabhjains:
> 
> Another related question is when the initramfs update fails, shall we un-register fadump i.e. write 0 to /sys/kernel/fadump/registered? If we do it this way, then "kdumpctl status" also work as expected.

Yeah that will bring consistency. We will have similar behaviour for kdump and fadump.
And as you said no changes will needed to `kdumpctl status`.

I will update the pull request. Thanks for the suggestion @coiby

sourabhjains:
> > Another related question is when the initramfs update fails, shall we un-register fadump i.e. write 0 to /sys/kernel/fadump/registered? If we do it this way, then "kdumpctl status" also work as expected.
> 
> Yeah that will bring consistency. We will have similar behaviour for kdump and fadump. And as you said no changes will needed to `kdumpctl status`.
> 
> I will update the pull request. Thanks for the suggestion @coiby

But what if user manually do:
echo 1 > /sys/kernel/fadump/register

We will have same problem again.

So I think for fadump the `kdumpctl status` should relay on `kdump service status`. Because as per the man kdumpctl:
```
 status Prints the current status of kdump service.  It returns  a  non-
              zero value if kdump is not operational. 
```

sourabhjains:
Any review/comment on this pull request?

coiby:
Hi @sourabhjains 

Thanks for the reminder! Sorry I missed your early message.

> Active kdump service guaranties that fadump was registered and iniitrd was built with capture service.

This is a good idea to me!

sourabhjains:
@coiby Are the changes in this pull request good to go, or would you prefer any further changes?

coiby:
> @coiby Are the changes in this pull request good to go, or would you prefer any further changes?

Thanks for the reminder! I thought you are going push a new change to check the initrd contains the capture if fadump initrd con reflect that "active kdump service also guaranties iniitrd was built with capture service".

But on second look, I think active kdump service has implied capture service has been included. So there is no need to check capture service.

Another question that comes to my mind when the kernel registers fadump during early boot, the old/outdated initramfs will be used and I expect users to interpret it as kdump is still operational . When "kdumpctl status" returns non-zero value when old initramfs is used but kdump.service fails, it will contradict the man page, 
> status 
> Prints the current status of kdump service.  It returns  a  non-zero value if kdump is not operational.

So I still think fadump shall be unregistered when kdump service fails.

As for your early concern,

> But what if user manually do:
>
> echo 1 > /sys/kernel/fadump/register
> 
> We will have same problem again.

My understanding writing 1 to the fadump register API is to use whatever existent initramfs. So  it doesn't seem to be a problem if an user uses this API while bypassing kdumpctl, or do I miss something?

PR #149, opened 2026-03-20
kdumpctl: Deprecate raw dump targets

### LINKED ISSUE #141: Remove the support for raw targets
When using a `raw` target, the dump is written directly to a disk/partition without using any file system. Support for `raw` targets were added with ea179a2 ("adding kexec frontend") in 2006, shortly before `makedumpfile` support was added. Unfortunately there is no documentation why the `raw` target was introduced.

My best guess is that at that time RAM was sparse and expensive. So not using a file system allowed to save  valuable RAM by reducing the crashkernel memory. In addition, `makedumpfile` didn't exist back then. So the full `/proc/vmcore`, which has a quite predictable size, was written to disk. So all in all the trade-off between saving RAM vs. wasting disk space was in favor of using `raw` targets.

But nowadays the world is different. For one, RAM is quite cheap. So saving a few MB of RAM is less important. In addition `makedumpfile` reduces the dump size quite a lot (compared to the total memory of the system). But this also makes the dump size less predictable. Meaning you need to add a larger safety to the disk/partition used as target. Otherwise you would risk to have an incomplete dump when the target runs out of space. So today, the trade-off no longer favors the use of 'raw' dump targets.

All in all, I don't think it makes sense to keep the `raw` dump target. Dropping support for it would allow us to remove some special handling in the code, thus simplifying the code and ultimately reduce the maintenance cost. In addition it would allow the removal of `is_raw_dump_target`, a wrapper of `kdump_get_conf_val`, and thus help with #139.

licliu:
The raw dump target (using dd to a partition) is being phased out in favor of filesystem-based targets. This patch adds deprecation warnings to the documentation and runtime scripts to alert users that support will be removed in a future release.

Due to editor settings, this patch also removes some trailing spaces.

Resolves: #141

Assisted-by: Google Gemini

### DIFF
--- a/gen-kdump-conf.sh
+++ b/gen-kdump-conf.sh
@@ -27,9 +27,9 @@ generate()
 #             overwritten to the default crahskernel value.
 #
 # raw <partition>
-#           - Will dd /proc/vmcore into <partition>.
-#             Use persistent device names for partition devices,
-#             such as /dev/vg/<devname>.
+#           - The 'raw' dump target is deprecated and will be removed in a future
+#             release. Please migrate to a filesystem-based dump target (e.g., ext4,
+#             xfs, nfs, or ssh).
 #
 # nfs <nfs mount>
 #           - Will mount nfs to <mnt>, and copy /proc/vmcore to
@@ -178,7 +178,6 @@ generate()
 #             (this option is mandatory to enable fence_kdump).
 #
 
-#raw /dev/vg/lv_kdump
 #ext4 /dev/vg/lv_kdump
 #ext4 LABEL=/boot
 #ext4 UUID=03138356-5e61-4ab3-b58e-27507ac41937
--- a/kdump.conf.5
+++ b/kdump.conf.5
@@ -3,7 +3,7 @@
 .SH NAME
 kdump.conf \- configuration file for kdump kernel.
 
-.SH DESCRIPTION 
+.SH DESCRIPTION
 
 kdump.conf is a configuration file for the kdump kernel crash
 collection service.
@@ -16,7 +16,7 @@ effect, restart the kdump service to rebuild the initrd.
 For most configurations, you can simply review the examples provided
 in the stock /etc/kdump.conf.
 
-.B NOTE: 
+.B NOTE:
 For filesystem dumps the dump target must be mounted before building
 kdump initramfs.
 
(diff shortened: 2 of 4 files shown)


gen-kdump-conf.sh
```diff
@@ -30,6 +30,7 @@ generate()
 #           - Will dd /proc/vmcore into <partition>.
 #             Use persistent device names for partition devices,
 #             such as /dev/vg/<devname>.
+#             NOTE: raw dump targets will be deprecated in a future version.
```
prudo1: Personally I'd remove everything following the dash `-` and replace it with what Gemeni has suggested. So the entry reads
```
# raw <partition>                                                               
#       The 'raw' dump target is deprecated and will be removed in a future        
#       release. Please migrate to a filesystem-based dump target (e.g., ext4,  
#       xfs, nfs, or ssh).
```

In addition I think the example below, aka. the line `#raw /dev/vg/lv_kdump`, should be removed as well.

kdump.conf.5
```diff
@@ -53,13 +53,14 @@ Also see kdumpctl(8).
 .B raw <partition>
 .RS
 Will dd /proc/vmcore into <partition>.  Use persistent device names for
-partition devices, such as /dev/vg/<devname>.
+partition devices, such as /dev/vg/<devname>. Raw dump targets will be
+deprecated in a future version.
 .RE
```
prudo1, line 59: Same like above in `gen-kdump-conf.sh`

kdumpctl
```diff
@@ -350,6 +350,7 @@ parse_config()
 			fi
 			;;
 		raw)
+			dwarn "Raw targets will be deprecated and removed in the future. Please migrate to a supported dump target."
```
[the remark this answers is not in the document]
prudo1: I find the suggestion made by Gemeni better. Can you please use it.

kexec-kdump-howto.txt
```diff
@@ -451,6 +451,8 @@ onto partition /dev/vg/lv_kdump. Restart the kdump service via
 initrd. Dump target should be persistent device name, such as lvm or device
 mapper canonical name.
 
+Raw targets will be deprecated in a future version.
+
 2) Dedicated file system
```
prudo1, line 456: Just for completion. Ideally we would use the same phrasing like in `gen-kdump-conf.sh` and the man page. But that would require to adjust the `Dedicated file system` section as well as it references this section. Personally I think that is too much for the deprecation and should be done when we remove the code. At that point we'll need to make more adjustments to `kexec-kdump-howto.txt` anyway. So I'm fine with keeping it this way.

prudo1 (changes requested):
Hi @licliu,

thanks for your PR. All in all it looks good to me. Can you please address the nits mentioned in the comments below.

In addition, I would remove `Resolves: #141` from the description. IMHO this PR is only the first step to resolve #141. For #141 to be fully resolved we would need to remove the code for raw targets.

Thanks
Philipp

licliu:
Thanks @prudo1 for carefully review! I will fix them in next version.

coiby (approved):

PR #161, opened 2026-08-03
99-kdump.conf: Add clevis-pin-file to omit_dracutmodules

rajucheerla:
Omit the clevis-pin-file module to avoid dracut warning during kdump initramfs rebuild.

 dracut[E]: Module 'clevis-pin-file' depends on module 'clevis',
 which can't be installed


coiby (approved):
Hi Raju,

Good to see you here! And thanks for sending this PR!

PR #163, opened 2026-08-21
kdumpctl: sync /etc/kernel/cmdline when resetting crashkernel

Vrushali-Z:
## Summary
After a single-kernel `grubby --update-kernel` crashkernel change, also
update `/etc/kernel/cmdline`. That stops the reset warning from firing
on every kernel update.
Does not use `--update-kernel=ALL`, so other kernels keep their own
crashkernel values.
## Jira
https://redhat.atlassian.net/browse/RHEL-188906
## Test plan
- [ ] Non-default crashkernel + `dnf reinstall kernel-core`: cmdline matches the default after first reset
- [ ] Second reinstall: no kdump reset warning
- [ ] Other installed kernels keep their original crashkernel


Vrushali-Z:
Hi maintainers, this is my first PR on this repo, so CI has not actually run yet.

Could someone please:
1. Approve and run GitHub Actions workflows
2. Comment `/packit build` so Packit can run (it currently says I cannot run tests/builds internally)

Jenkins `pr-merge` failed with qemu-img "Image creation needs a size parameter" while starting kola VMs. That looks like the same CoreOS CI issue as on other open PRs, not this crashkernel change.

Thanks!

prudo1:
FYI, there is an ongoing discussion in the referenced Jira issue

coiby:
Hi @Vrushali-Z,

Different kernels can have different crashkernel values because the default crashkernel value is not static. For example, 4K and 64K kernels have different crashkernel value. For more details, you can check the `kdump_get_arch_recommend_crashkernel` function in kdump-lib.sh. So I don't think it's a good idea to update /etc/kernel/cmdline.

Who emits the reset warning and what's the warning like?

Vrushali-Z:
> Hi @Vrushali-Z,
> 
> Different kernels can have different crashkernel values because the default crashkernel value is not static. For example, 4K and 64K kernels have different crashkernel value. For more details, you can check the `kdump_get_arch_recommend_crashkernel` function in kdump-lib.sh. So I don't think it's a good idea to update /etc/kernel/cmdline.
> 
> Who emits the reset warning and what's the warning like?


Hi @coiby,
Thanks for the review.
The message comes from `kdumpctl _update_crashkernel()`, called by `92-crashkernel.install` during `kernel-install` (`kdumpctl _reset-crashkernel-for-installed_kernel`, when `auto_reset_crashkernel` is not `no` and kdump is enabled). It looks like this:
kdump: For kernel=/boot/vmlinuz-<version>, crashkernel=2G-64G:256M,64G-:512M now. Please reboot the system for the change to take effect. Note if you don't want kdump-utils to manage the crashkernel kernel parameter, please set auto_reset_crashkernel=no in /etc/kdump.conf.
You are right that writing a single crashkernel value into `/etc/kernel/cmdline` is not a good idea. That file is copied into new BLS entries, and different kernels can need different defaults (for example 4K vs 64K on aarch64, via `kdump_get_arch_recommend_crashkernel`).
I also saw Philipp’s comments on RHEL-188906: `20-grub.install` reapplies `/etc/kernel/cmdline` first, then `92-crashkernel.install` resets the BLS entry, so the message on every kernel (re)install is expected when those two sources disagree. kdumpctl should not silently overwrite `/etc/kernel/cmdline`.
Should I close this PR, or would you like a different follow-up?

